PR #1, opened 2013-06-23
Hint at fixes for PySide recipe

ChrisBeaumont:
(This PR is just to discuss an issue with the PySide recipe; it shouldn't be accepted)

PySide includes 3 binary programs and a supplementary python module that aren't currently pulled into anaconda. They are

 * pyside-uic and the pysideuic module, which compile UI xml files into python
 * pyside-rcc, which compiles external resources (like icon files) into python modules
 * pyside-lupdate, which handles internationalization

I've hinted at how to pull these into anaconda on OSX, but I'm mostly guessing (I don't know how to recreate the conda build process)

cc @asmeurer


msarahan:
As I understand it, pyqt is strongly preferred to pyside. Closing this PR. Reopen if you'd like.

PR #108, opened 2014-04-18
Openturns

dubourg:
This is a conda-recipe for OpenTURNS (1.3). See http://openturns.org.
The Linux version compiles it from source fetched from the latest tagged version on the subversion repos (some dependencies are not available in conda, so I mentionned them as comments in build.sh).
The Windows version uses flags on the latest nsis binary installer fetched from sourceforge.


dubourg:
Closing this... See #235 instead!

PR #141, opened 2014-07-11
freeimage 3.16.0 builds on osx 10.8.5 / llvm 5.1

jf---:
As on the tin...
builds [libfreeimage](http://freeimage.sourceforge.net/), on osx.
untested on linux
freeimage is a dependency for [OpenCASCADE / OCE](opencascade.org), another PR for that one coming on short notice.


almarklein:
I think this can be closed per #362. If the current recipe does not work, it should probably get a new PR against the latest master. cc @msarahan

msarahan:
Agreed, thanks for pointing it out.

PR #144, opened 2014-07-14
Improve matplotlib package.

mdboom:
This improves the matplotlib package in a few ways:

- It uses `pkg-config` to find its dependencies (as discussed with @asmeurer at Scipy sprints), which allows for the removing of some unnecessary patches.

- It removes the test data from the main distribution -- about 50MB, and creates a new recipe to install the test data if the user wants it.

First Conda package, so let me know how it can be improved...


mdboom:
Superceded by #440

PR #162, opened 2014-08-13
recipes for Qt5/sip/PyQt5 stack on windows, linux, osx

ddale:
Compiles Qt5, sip, and pyqt5 using MSVC Express (tested with 2008 for py27, 2010 for py3k). Uses precompiled ICU binaries (build and runtime requirement for Qt5 on windows), and uses precompiled jom executable to perform the build in parallel.

Qt5 is given a python requirement to differentiate qt5 built with msvc2008 for py27 and qt5 built with msvc2010 for py3k. Linking pyqt5 built with msvc2008 to a qt5 built with msvc2010 resulted in crashes in a few of PyQt5's examples, notably qtdemo.py

### DIFF
--- a/pyqt/notes.md
+++ b/pyqt/notes.md
@@ -1,0 +1,8 @@
+== Windows ==
+
+Building the pyqt conda package for windows requires special preparation. The
+recipe assumes you have run the executables from Riverbank Computing and
+installed into `C:\staging`. This directory should not exist before installing
+to that location, i.e., it should not be a working python environment. The
+pyqt recipe's bld.bat just copies the contents of `C:\staging` into our conda
+build environment.
--- a/pyqt/notes.md
+++ b/pyqt/notes.md
@@ -1,8 +1,24 @@
 == Windows ==
 
-Building the pyqt conda package for windows requires special preparation. The
-recipe assumes you have run the executables from Riverbank Computing and
-installed into `C:\staging`. This directory should not exist before installing
-to that location, i.e., it should not be a working python environment. The
-pyqt recipe's bld.bat just copies the contents of `C:\staging` into our conda
-build environment.
+Building the pyqt conda package for windows requires special
+preparation. The recipe assumes you have run the executables from
+Riverbank Computing and installed into `C:\staging`. This directory
+should not exist before installing to that location, i.e., it should
+not be a working python environment. The pyqt recipe's bld.bat just
+copies the contents of `C:\staging` into our conda build environment.
+
+An alternative approach would be to add a Qt(4) package for windows,
+but then we would need to patch either Qt4 or Qt5 to prevent
+collisions of executables (like 'designer' and 'qmake', the DLLs at
+least do not collide), and also PyQt4 or PyQt5 would have to be
+patched to find the renamed executables.
+
+Another approach to pyqt would be to build PyQt4 against Qt5 for
+python3. This would allow building PyQt4 from source on windows, but
+based on the documentation at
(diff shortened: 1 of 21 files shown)


pyqt/notes.md
```diff
@@ -0,0 +1,24 @@
+== Windows ==
+
+Building the pyqt conda package for windows requires special
+preparation. The recipe assumes you have run the executables from
+Riverbank Computing and installed into `C:\staging`. This directory
```
ccordoba12: Could `C:\staging` be changed to a directory inside the user profile?

qt5/README.md
```diff
@@ -0,0 +1,6 @@
+qt5-conda-recipe
+================
+
+Strategy for a qt5 conda package that can be installed alongside qt(4), largely taken from http://www.linuxfromscratch.org/blfs/view/7.4/x/qt5.html , installing executables to `$PREFIX/qt5/bin` and making links in `$PREFIX/bin` with names like `designer-qt5`. (A different approach is now advocated at http://www.linuxfromscratch.org/blfs/view/7.5/x/qt5.html, where a shell script is used to switch between qt4 and qt5 by modifying the `PATH`.)
```
ccordoba12: Could you break this paragraph in several lines? It's just one big line right now :-)

ccordoba12:
@ddale Although I see that you've put a lot of effort provinding a conda recipe for Windows, could it be possible to install from a binary in that platform?

I'm just exploring the possibilities because we (at Spyder) are going to need support for Qt5 in the future to embed the IPython notebook :-)

ccordoba12:
In case that's not possible, could you post here the patch for `conda-build` to make this work for Windows? I'll try to setup a virtual machine and run your recipes this weekend :-)

ddale:
You can build qt5 on Windows with the existing conda build, I think.

ccordoba12:
Great! I'll give it a chance and let you know what happens.

jdreaver:
I don't know if anyone has encountered this issue, but when I tried to build the pyqt5 recipe as-is, I run out of memory. I see that build is configured to run `$CPU_COUNT` jobs, and my laptop has 8 cores. It also has 16GB of RAM. If I set the number of jobs to 1, it seems to work fine.

Also, thanks @ddale for your hard work! I really want to use PyQt5, but my company does continuous integration on Linux and Windows, and there is no way we can wait for PyQt5 to compile every time the test suite runs. A conda build seems to be the only way.

ccordoba12:
@jdreaver thanks for the info. @ddale is busy right now, so I'm going to pick up his recipes soon. That's why I'd like to ask you: do you have time to test the other recipes here? are they working for you?

Unfortunately I'm busy right now fixing things for Qt4, but if I received a green light from you, I'd merge @ddale work and fix errors on master.

jdreaver:
I don't have time right now unfortunately :/ I will have more time at the beginning of March.

The PyQt5 recipe didn't seem to work on Arch Linux, 64 bit. I got an error relating to `xcb`. I will have to investigate when I get more time.

asmeurer:
I had several issues with these recipes when I was testing them. As far as I know, not all issues have been fixed.

ccordoba12:
Thanks @asmeurer, I'll cherry pick @ddale's work then and open a new PR.

asmeurer:
If you're going to work on it feel free to just merge this and work off master.

ccordoba12:
Yes, you're right. Since I need @ddale changes to sip and PyQt4 recipes, I think it's better to merge now.

PR #186, opened 2014-10-05
Add pims and trackpy

danielballan:
Here are working recipes for [pims](https://github.com/soft-matter/pims) and [trackpy](https://github.com/soft-matter/trackpy). New versions of both of these were uploaded to PyPI on Friday.


danielballan:
Thanks!

PR #192, opened 2014-10-07
mistune

hhuuggoo:


mistune/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+
+  # If this is a new build for the same version, increment the build
+  # number. If you do not include this key, it defaults to 0.
+  # number: 1
+
+requirements:
+  build:
+    - python
+    - setuptools
```
asmeurer: @ilanschnell we should probably do this.

ilanschnell: Yeah, I'm doing that right now

PR #195, opened 2014-10-13
Add recipe for tifffile.

danielballan:


tifffile/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+requirements:
+  build:
+    - python
+    - numpy
+
+  run:
+    - python
+    - numpy
+
```
stuarteberg: The docstring in `tifffile.py` indicates [specific version requirements](https://github.com/blink1073/tifffile/blob/master/tifffile.py#L65):
  - `python 2.7*|>=3.4`
  - `numpy >=1.8.2`

jakirkham: @stuarteberg, adding that python constraint will cause the build to fail.

stuarteberg: Seems to work just file in my own recipe repo:
https://github.com/ilastik/ilastik-build-conda/blob/master/tifffile/meta.yaml

jakirkham: Well, the import test failed in my case and it appeared to try and use Python 3.4 even though I am using the Python 2.7 based miniconda.

stuarteberg: Strange.  Well, it works for me, and, as far as I can tell, it obeys the conda recipe [package match specification syntax](http://conda.pydata.org/docs/spec.html#package-match-specifications). Are you able to use the recipe I linked to above?  For that matter are you able to use my binary via `conda install -c ilastik tifffile`?

jakirkham: I literally copied what you wrote above.

jakirkham: Yeah, then I tried your recipe verbatim. Maybe my conda install is messed up. :/

danielballan: Weird, the suggested change also causes the build to fail for me. I get a gcc error on OSX, only when these versions are pinned. Building on 2.7, 3.4 with numpy 1.10 works fine as long as the recipe does not actually pin them.

PR #213, opened 2014-12-10
Update IPython recipe to current stable version

takluyver:
Also, setuptools is no longer a build-dep on Python 3


takluyver:
Heads up, the stable version when I made this PR is no longer the latest - currently we're at 3.1.0.

PR #214, opened 2014-12-11
bump cloog to 0.18.1

minrk:
Avoids impossible to complete dependency in gcc-4.9, which depends on 0.18.1, but would build 0.18.0 from the recipe, over and over again, forever.


asmeurer:
I haven't been able to get the 4.9 recipe to work. I've been using 4.8, which depends on this version of cloog (the version restrictions are based on the gcc docs).

PR #217, opened 2014-12-14
update wxpython recipe to 3.0.2.0

aaren:
Tested on Debian Jessie. 

Fixed a typo in the `CXXFLAGS`. Found that I had to set `CPPFLAGS` as well.
Also I needed to set `-std=c++11` in the c++ flags - this might be Debian
specific.

I had to create a symlink for `g++44` because it was getting called in the
build for some reason:

    ln -s /usr/bin/g++ /usr/bin/g++44

I'm not sure why this is. I tried grepping the source for 'g++44' and variants
but couldn't find anything (my g++ is 4.9 and I don't have 4.4 available).

My build is on binstar: https://binstar.org/aaren/wxpython


ilanschnell:
Cool, thanks a lot

PR #223, opened 2014-12-20
ncurses: use terminfo databases from host system by default

immerrr:
Terminal devices (incl. virtual) are generally provided by host system
and terminfo db should describe their capabilities.  It is more likely
that host system databases will fit the devices provided by the host system
than the databases shipped with conda.

In this respect, my previous effort, #177, is incorrect.


immerrr:
I'm going to close this, haven't had problem with libtinfo/libncurses for a while.

PR #227, opened 2014-12-24
Use a tagged version of ggplot (from pypi) instead of 'master'

derickl:
I think this will make it easier to release updated recipes

### DIFF
--- a/ggplot/meta.yaml
+++ b/ggplot/meta.yaml
@@ -1,11 +1,12 @@
 
 package:
     name: ggplot
-    version: "master"
+    version: !!str 0.6.5
 
 source:
-    git_url: https://github.com/yhat/ggplot
-    git_tag: "master"
+    fn: ggplot-0.6.5.tar.gz
+    url: https://pypi.python.org/packages/source/g/ggplot/ggplot-0.6.5.tar.gz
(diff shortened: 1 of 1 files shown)


ggplot/meta.yaml
```diff
@@ -1,11 +1,12 @@
 
 package:
     name: ggplot
-    version: "master"
+    version: !!str 0.6.5
```
ccordoba12: Please change this to 0.6.6 and remove the `!!str` part

ggplot/meta.yaml
```diff
@@ -1,11 +1,12 @@
...
 
 source:
-    git_url: https://github.com/yhat/ggplot
-    git_tag: "master"
+    fn: ggplot-0.6.5.tar.gz
+    url: https://pypi.python.org/packages/source/g/ggplot/ggplot-0.6.5.tar.gz
```
ccordoba12: Update to 0.6.6 here too

ggplot/meta.yaml
```diff
@@ -1,11 +1,12 @@
...
+    md5: c825f9ca48bc9b12c2e2517c7f8b9538
```
ccordoba12: Don't forget updating this one too :-)

PR #235, opened 2015-01-12
Add openturns package

jschueller:
Tested on linux64, win32, win64.

On windows, depends on:
https://github.com/conda/conda-build/pull/295

### DIFF
--- a/lapack/build.sh
+++ b/lapack/build.sh
@@ -1,0 +1,11 @@
+mkdir build
+cd build
+
+cmake \
+    -DCMAKE_INSTALL_PREFIX=$PREFIX \
+    -DBUILD_TESTING=OFF \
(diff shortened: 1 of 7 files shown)


dubourg:
Tested both win-32 and linux-64 binaries from [binstar](https://binstar.org/openturns/openturns).
Both work, although the linux-64 package seems to have a broken path specification to the openturns.conf file:
```python
>>> import openturns
WRN - The configuration file has not been found, using default parameters.
```
Indeed, it is located at `$ANACONDA_PATH/etc/openturns/openturns.conf` and openturns can't find it. Shouldn't it be `$ANACONDA_PATH/etc/openturns.conf`.

jschueller:
just fixed the config file not taken into account at 427c2e5, as well as the tbb and hmat dependencies.

can you "conda update openturns" ?

PR #238, opened 2015-01-14
Update yt recipe for yt 3.1.0

ngoldbaum:
You'll likely see some test failures related to `yt.utilities.lib.ragged_arrays` when the unit tests run.  This is a known issue and can be ignored.

### DIFF
--- a/yt/meta.yaml
+++ b/yt/meta.yaml
@@ -1,11 +1,11 @@
 package:
   name: yt
-  version: 3.0.2
+  version: 3.1.0
 
 source:
-  fn: yt-3.0.2.tar.gz
-  url: https://pypi.python.org/packages/source/y/yt/yt-3.0.2.tar.gz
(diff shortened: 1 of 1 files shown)


asmeurer:
@ilanschnell you should probably update this in Anaconda. Did you ever fix the 32-bit support?

ngoldbaum:
32 bit support is fixed.  You'll see the same test failures I talked about in the PR description, but you can safely ignore it.

ilanschnell:
I'm updating to yt 3.1.0 right now.  32-bit is compiling now as well.

asmeurer:
What about Python 3?

ngoldbaum:
No python 3 yet, hopefully we'll have py3 support in the next major version.

You can take a look at the changelog here: http://yt-project.org/docs/3.1/reference/changelog.html#version-3-1

PR #247, opened 2015-01-30
Fix SimpleITK build script for windows

blowekamp:
Update SimpleITK hash, with CMake build fixes for EXAMPLES, and an escape
sequence issue.

Use internal third party libraries on windows since depencent dll aren't automatically
include in the path at run-time. See:
https://groups.google.com/a/continuum.io/forum/#!topic/conda/wnRbABPq4hE

Use CONDA_PY for python version, as PY_VER is not being defined.

### DIFF
--- a/simpleitk/bld.bat
+++ b/simpleitk/bld.bat
@@ -1,15 +1,18 @@
-mkdir C:\bld
-cd C:\bld
+mkdir C:\b
+cd C:\b
 
 if "%ARCH%" == "32" set "CMAKE_GENERATOR=Visual Studio 9 2008"
-if "%ARCH%" == "64" set "CMAKE_GENERATOR=Visual Studio 9 Win64"
+if "%ARCH%" == "64" set "CMAKE_GENERATOR=Visual Studio 9 2008 Win64"
+
+REM Remove dot from PY_VER for use in library name
+set MY_PY_VER=%PY_VER:.=%
 
 REM Configure Step
 cmake -G "%CMAKE_GENERATOR%" ^
     -D SimpleITK_BUILD_DISTRIBUTE:BOOL=ON ^
-    -D SimpleITK_BUILD_STRIP:BOOL=ON ^
     -D BUILD_SHARED_LIBS:BOOL=OFF ^
     -D BUILD_TESTING:BOOL=OFF ^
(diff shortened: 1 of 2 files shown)


asmeurer:
PY_VER not being defined sounds like a bug in conda-build. You shouldn't rely on CONDA_PY being defined as I don't think it is if you use `conda build --py 2.7`.

blowekamp:
OK, I'll revert that change. I  "set CONDA_PY=34" then "conda build simpleitk" and PY_VER was not set. Resulting in a linking error.

Related on linux/osx I am setting the python library to:
"PYTHON_LIBRARY:FILEPATH=$PREFIX/lib/libpython${MY_PY_VER}.${SO_EXT}" 

Where MY_PY_VER adds the "m" character suffix for python 3. I don't recall what this library suffix means but it might be good to have it accessible in the build environment somehow.

asmeurer:
From my reading of the conda-build code `PY_VER` should always be set. I just made a test recipe and it was set in there.

asmeurer:
Oh this is Windows. I wonder if that is related. Let me check.

asmeurer:
No, it works for me in Windows too.

asmeurer:
I see the problem. It is `libpython33.lib`, but `PY_VER` is `3.3` (with the `.`).

blowekamp:
That sounds right. I didn't recall that correctly away from the computer. 

BTW, the build that was hanging at scipy was caused by ez_setup using an old version of curl or wget on rh5 which hung with the requested http ssl connection. No longer using ez setup as suggested.

asmeurer:
I don't remember the build hanging at scipy. It just took a really long time, and too much memory for my VM.

blowekamp:
OK, Is there a suggestions for this line:

 -D "PYTHON_LIBRARY:FILEPATH=%PREFIX%/libs/python%CONDA_PY%.lib" ^

Or a robust where to specify the python library to use?  I think I can find an arcane batch command to remove the "." for PY_VER, just  checking if there is a better recommendation.

Thanks.

asmeurer:
I think `%PY_VER:.=% will remove it. See http://www.dostips.com/DtTipsStringManipulation.php#Snippets.Replace

blowekamp:
Thank you that did the trick.

PR #249, opened 2015-02-02
Upgraded prettyplotlib to v1.7

ovz:


ccordoba12:
I think prettyplotlib is unmaintained now, right?

ovz:
I still have ipython notebooks that use it. Would be convenient to keep last stable version as a conda package. WBR Oleg Zhylin ovz@yahoo.com 

     On Monday, February 2, 2015 12:17 PM, Carlos Cordoba <notifications@github.com> wrote:
   

 I think prettyplotlib is unmaintained now, right?—

ccordoba12:
If prettyplotlib is pip installable and is a pure Python package, I'd be in favor of removing it from conda-recipes.

@asmeurer, @olgabot what do you think?

ccordoba12:
The package is no longer maintained (as @olgabot mentioned), and it's easily pip-installable (as I already verified).

Now that pip is part of every conda environment, I don't see the need to maintain this recipe as part of conda-recipes.

asmeurer:
It's useful to have recipes for pure Python packages (most recipes here are), but I agree we don't need to have recipes for dead projects.

ccordoba12:
@asmeurer, I understand. My main reason was really that the project is dead now, but easily installable by other means.

PR #250, opened 2015-02-02
Updates to HDF4

jgomezdans:
This PR does two things: Firstly, it updates the HDF4 version to 4.2.10, and secondly, it integrates a patch to increase the number of simultaneously opened files to 2048. This is a benign patch (we have experienced no problems with deploying this patch on lots of systems), and it adds some important extra functionality.


jgomezdans:
Please test this PR, as I can get it to work fine in one of my computers, whereas it is failing on another. I presume the patch is somehow bombing?

PR #251, opened 2015-02-02
Updates HDF4 to version 4.2.10

jgomezdans:
It is tested on Linux on X86..._64, and works file. **Additionally**, it introduces a patch to **increase the maximum allowed simultaneously opened files** from 32 to 2048. This patch is beningn, in the sense that we have deployed the binaries on many systems where other users use the library and have found no side effects. This allows other packages like GDAL to exploit the usefulness of having many possible HDF4 files opened simultaneously (e.g. see [here](http://www.gdal.org/frmt_hdf4.html)])

### DIFF
--- a/hdf4/build.sh
+++ b/hdf4/build.sh
@@ -4,32 +4,14 @@ mkdir -vp ${PREFIX}/bin;
 
 ARCH="$(uname 2>/dev/null)"
 
-export CFLAGS="-m64 -pipe -O2 -march=x86-64 -fPIC"
-export CXXFLAGS="${CFLAGS}"
-#export CPPFLAGS="-I${PREFIX}/include"
-#export LDFLAGS="-L${PREFIX}/lib"
-
-LinuxInstallation() {
-
-    chmod +x configure;
-
-    ./configure \
-        --disable-static \
-        --enable-linux-lfs \
-        --with-ssl \
-        --with-zlib \
-        --prefix=${PREFIX} || return 1;
-    make || return 1;
-    make install || return 1;
-
-    rm -rf ${PREFIX}/share/hdf4_examples;
-
-    return 0;
-}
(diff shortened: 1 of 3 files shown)


WeatherGod:
Just had a project start that had need for hdf4 support, so now I have some cycles to examine this PR. It looks like the hdf group just updated the package to v4.2.11, so the link to 4.2.10 now fails because of their directory structure.

WeatherGod:
To continue on, I simply tweeked meta.yaml to point to the "previous" directory instead of "HDF_Current". Now I get an error when applying the patch:
```
Applying patch: u'/network/apps/conda-recipes/hdf4/max_files.patch'
can't find file to patch at input line 4
Perhaps you used the wrong -p or --strip option?
The text leading up to this was:
--------------------------
|diff -rupN hdf-4.2.10/hdf/src/hlimits.h new/hdf/src/hlimits.h
|--- hdf-4.2.10/hdf/src/hlimits.h	2014-02-10 02:28:49.000000000 +0000
|+++ new/hdf/src/hlimits.h	2014-09-02 14:07:48.015297738 +0100
--------------------------
```

WeatherGod:
looks like there is a build requirement for yacc...

WeatherGod:
heh, looks like the machine I was testing on was really stripped down to the bare minimums. I'll test on a different machine.

WeatherGod:
ok, it seems to build fine once I bypass the bad link and the patch. I want to test it with your other PR, but it seems to want to keep getting hdf4 from the conda channel on binstar (where there is none) rather than using the hdf4 that I installed directly. Is there something I missing that needs to be set?

WeatherGod:
heh, nm. Don't know what I was doing wrong, but pyhdf will now build and install properly.

PR #252, opened 2015-02-02
Updates GDAL to 1.11.1, and adds HDF4, HDF5, LibnetCDF formats, an...

jgomezdans:
Updated GDAL to 1.11.1, but also added HDF4, HDF5, LibnetCDF formats, and dependencies. Added a short script to test opening MODIS HDF files (a format often requested by users), as well as listing the supported formats, both for vector and raster. Only tested on Linux X86, should be easy to extend to MacOSX. This patch would make the package extremely useful for users of remote sensing data, as it now includes a wealth of often-used formats.

### DIFF
--- a/gdal/build.sh
+++ b/gdal/build.sh
@@ -6,9 +6,8 @@
 --with-hdf4=$PREFIX \
 --with-netcdf=$PREFIX \
 --with-xerces=$PREFIX \
---with-sqlite=$PREFIX \
+--with-sqlite3=$PREFIX \
 --without-pam \
---with-python \
 --disable-rpath
 
 make 
--- a/gdal/build.sh
+++ b/gdal/build.sh
@@ -1,7 +1,22 @@
 #!/bin/bash
+./configure --with-python --prefix=$PREFIX \
+--with-geos=$PREFIX/bin/geos-config \
+--with-static-proj4=$PREFIX \
+--with-hdf5=$PREFIX \
+--with-hdf4=$PREFIX \
+--with-netcdf=$PREFIX \
+--with-xerces=$PREFIX \
+--with-sqlite=$PREFIX \
+--without-pam \
+--with-python \
+--disable-rpath
 
-bash configure --with-python --prefix=$PREFIX
-make
+make 
 make install
 
-rm -rf $PREFIX/share
+# Copy data files 
+mkdir -p $PREFIX/share/gdal/
+cp data/*csv $PREFIX/share/gdal/
+cp data/*wkt $PREFIX/share/gdal/
+
+
--- a/gdal/build.sh
+++ b/gdal/build.sh
@@ -7,6 +7,7 @@
 --with-netcdf=$PREFIX \
 --with-xerces=$PREFIX \
 --with-sqlite3=$PREFIX \
+--with-curl=$PREFIX/bin/curl-config \
 --without-pam \
 --disable-rpath
 
--- a/gdal/meta.yaml
+++ b/gdal/meta.yaml
@@ -1,18 +1,29 @@
 package:
   name: gdal
-  version: 1.10.1
+  version: 1.11.1
 
 source:
-  fn: gdal-1.10.1.tar.gz
-  url: http://download.osgeo.org/gdal/1.10.1/gdal-1.10.1.tar.gz
-  md5: 86b2c71a910d826a7fe6ebb43a532fb7
+  fn: gdal-1.11.1.tar.gz
+  url: http://download.osgeo.org/gdal/1.11.1/gdal-1.11.1.tar.gz
+  md5: 7555f55855f613be49e6508eed0ac3fa
 
 requirements:
   build:
     - python
+    - hdf4
+    - hdf5 
+    - numpy >=1.9
+    - geos
+    - proj4
+    - curl
   run:
     - python
-
+    - hdf5
+    - hdf4
+    - numpy >=1.9
+    - geos
+    - proj4
+    - curl
 test:
   files:
     - os1_hw.py
--- a/gdal/meta.yaml
+++ b/gdal/meta.yaml
@@ -1,11 +1,11 @@
 package:
   name: gdal
-  version: 1.11.1
+  version: 1.11.2
 
 source:
-  fn: gdal-1.11.1.tar.gz
-  url: http://download.osgeo.org/gdal/1.11.1/gdal-1.11.1.tar.gz
-  md5: 7555f55855f613be49e6508eed0ac3fa
+  fn: gdal-1.11.2.tar.gz
+  url: http://download.osgeo.org/gdal/1.11.2/gdal-1.11.2.tar.gz
+  md5: 866a46f72b1feadd60310206439c1a76
 
 requirements:
   build:
--- a/gdal/meta.yaml
+++ b/gdal/meta.yaml
@@ -16,6 +16,7 @@ requirements:
     - geos
     - proj4
     - curl
+    - libnetcdf
   run:
     - python
     - hdf5
@@ -24,6 +25,7 @@ requirements:
     - geos
     - proj4
     - curl
+    - libnetcdf
 test:
   files:
     - os1_hw.py
--- a/gdal/meta.yaml
+++ b/gdal/meta.yaml
@@ -12,7 +12,7 @@ requirements:
     - python
     - hdf4
     - hdf5 
-    - numpy >=1.9
+    - numpy 
     - geos
     - proj4
     - curl
@@ -20,7 +20,7 @@ requirements:
     - python
     - hdf5
     - hdf4
-    - numpy >=1.9
+    - numpy
     - geos
     - proj4
     - curl
--- a/gdal/run_test.py
+++ b/gdal/run_test.py
@@ -7,9 +7,28 @@ import gdal
 import gdalconst
 import ogr
 import osr
+import urllib2
 
+print( "OGR Vector drivers")
+print("==================")
 cnt = ogr.GetDriverCount()
+for i in range(cnt):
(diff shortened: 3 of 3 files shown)


gdal/build.sh
```diff
@@ -1,7 +1,22 @@
...
+./configure --with-python --prefix=$PREFIX \
+--with-geos=$PREFIX/bin/geos-config \
+--with-static-proj4=$PREFIX \
+--with-hdf5=$PREFIX \
+--with-hdf4=$PREFIX \
+--with-netcdf=$PREFIX \
+--with-xerces=$PREFIX \
+--with-sqlite=$PREFIX \
+--without-pam \
+--with-python \
```
WeatherGod: duplicate. Also, you should really be doing `--with-python=${PYTHON}`, I think to guarantee that it is compiled against the right python.

gdal/meta.yaml
```diff
@@ -1,18 +1,29 @@
...
+  fn: gdal-1.11.1.tar.gz
+  url: http://download.osgeo.org/gdal/1.11.1/gdal-1.11.1.tar.gz
+  md5: 7555f55855f613be49e6508eed0ac3fa
 
 requirements:
   build:
     - python
+    - hdf4
+    - hdf5 
+    - numpy >=1.9
```
WeatherGod: what is the reason for this restriction?

gdal/run_test.py
```diff
@@ -7,9 +7,28 @@
...
 cnt = ogr.GetDriverCount()
+for i in range(cnt):
+    print(ogr.GetDriver(i).GetName())
+
+print ("GDAL Raster drivers")
+print ("===================")
+cnt = gdal.GetDriverCount()
 for i in xrange(cnt):
-    print ogr.GetDriver(i).GetName()
+    print (gdal.GetDriver(i).LongName)
```
WeatherGod: extraneous space between print and parens

gdal/run_test.py
```diff
@@ -7,9 +7,28 @@
...
+
+print ("Total number of vector drivers: %d" % ogr.GetDriverCount())
+print ("Total number of vector drivers: %d" % gdal.GetDriverCount())
+
+print ("Testing opening a MODIS file...")
+fp = open ( "MCD45A1.A2006213.h17v04.051.2013060204001.hdf", 'w')
```
WeatherGod: more extraneous spaces...

WeatherGod:
Not crazy about the extra test via the url download. Why not use whatever built-in tests that come with the tarball?

gdal/build.sh
```diff
@@ -1,7 +1,22 @@
 #!/bin/bash
+./configure --with-python --prefix=$PREFIX \
+--with-geos=$PREFIX/bin/geos-config \
+--with-static-proj4=$PREFIX \
+--with-hdf5=$PREFIX \
+--with-hdf4=$PREFIX \
+--with-netcdf=$PREFIX \
+--with-xerces=$PREFIX \
+--with-sqlite=$PREFIX \
```
WeatherGod: IIRC (I am not at my work computer where I was doing some experiments), --with-sqlite isn't an actual switch for configure.

jgomezdans: You're right, it's ``--with-sqlite3``

jgomezdans:
OK, taken most of your comments on board. A couple of things:

1. The ``--with--python``  flag basically tells the compilation to do the Python bindings, so no extra args are needed
2. The numpy version was required in our system, otherwise for some reason, conda was picking up numpy 1.8 (which is the one we have installed here still)
3. The download URL test isn't necessary, but tests whether the MODIS HDF files (which is what I use everyday) work fine. If they don't work, this is as good as useless to me (and many other users). Potentially, just testing against having the HDF4 driver would be equivalent.

gdal/build.sh
```diff
@@ -1,7 +1,21 @@
...
+./configure --with-python --prefix=$PREFIX \
+--with-geos=$PREFIX/bin/geos-config \
+--with-static-proj4=$PREFIX \
+--with-hdf5=$PREFIX \
+--with-hdf4=$PREFIX \
+--with-netcdf=$PREFIX \
+--with-xerces=$PREFIX \
+--with-sqlite3=$PREFIX \
+--without-pam \
+--disable-rpath
```
WeatherGod: Why disable rpath? The conda build system automatically patches the binaries to have relocatable paths. Without rpaths, the package isn't relocatable anymore.

WeatherGod: At least, I am pretty sure it isn't relocatable anymore. It is entirely possible that I am wrong about this. I won't be able to test this PR until tomorrow when I can get back to the office (I am a bit snowed in at the moment...)

jgomezdans: Other than "works for me", I can't really say ;-) I'd be very grateful if you could test this PR together with PR #251 too...

asmeurer: It probably doesn't matter if you use this flag or not, since conda chanages the binaries at the end using patchelf.

gdal/build.sh
```diff
@@ -1,7 +1,21 @@
 #!/bin/bash
+./configure --with-python --prefix=$PREFIX \
+--with-geos=$PREFIX/bin/geos-config \
+--with-static-proj4=$PREFIX \
+--with-hdf5=$PREFIX \
+--with-hdf4=$PREFIX \
+--with-netcdf=$PREFIX \
+--with-xerces=$PREFIX \
+--with-sqlite3=$PREFIX \
```
WeatherGod: since we are setting "curl" as a requirement, perhaps "--with-curl=$PREFIX/bin/curl-config" should be added?

gdal/meta.yaml
```diff
@@ -1,18 +1,29 @@
...
+    - curl
   run:
     - python
-
+    - hdf5
+    - hdf4
+    - numpy
+    - geos
+    - proj4
+    - curl
```
WeatherGod: netcdf4 package is missing

gdal/meta.yaml
```diff
@@ -1,18 +1,29 @@
...
   run:
     - python
-
+    - hdf5
+    - hdf4
+    - numpy
+    - geos
+    - proj4
```
WeatherGod: I don't see proj4 in the conda channel, so I get an error trying to build the package.

WeatherGod: Just took a quick peek to see what the conda recipe for basemap is (because it uses proj4 IIRC), and it doesn't seem like it exists in this repo!

gdal/build.sh
WeatherGod: I am not totally familiar with exactly how patchelf is used, but my understanding was that it only seeked out and modified rpaths. Once I finally get this package building on our system, I'll be able to try relocating it to some of the test machines to see if it is truly relocatable.

WeatherGod:
I was having issues completing a build of the package on the CentOS5 system I have. I kept getting undefined references for libz. I found that if I add a "--with-libz=internal" flag to the build, the build completes.

Seems to be having issues with hdf5, though:
```
ImportError: libhdf5_hl.so.7: cannot open shared object file: No such file or directory
```
Does this mean that I need to build and install the hdf5 package using your other PR first? I wasn't sure if the changes there were relevant to my linux system.

WeatherGod:
and just noticed that the PR was for hdf4, not hdf5, which I did not include for my tests (since your PR doesn't straight-out work for me yet due to the lack of proj4 and such). Will need to investigate further.

gdal/meta.yaml
jgomezdans: proj4 is available in this repo (see [here](https://github.com/conda/conda-recipes/tree/master/proj4) ). I do compile individual packages of all these dependencies to have latest versions (e.g. GEOS 3.4.2) or other patches that I need (eg #251)

jgomezdans:
HDF4 and HDF5 are both available in the main conda repo (HDF5 is used by e.g. pytables). #251 is a patch and update that's useful to have in HDF4 before you build GDAL against it. In my systems (RHEL6 and Ubuntu 14.10), I don't need the "--with-libz" option set up, but it wouldn't hurt having it stated explicitly.

gdal/meta.yaml
WeatherGod: Ah, weird, don't know how I missed that. So, why isn't the proj4 package available through the standard conda channel? What is the policy regarding conda-recipes that require packages not yet available through the standard conda channel?

WeatherGod:
Right, I know that the HDFs are available. For some reason, my build (which I kept very minimal in order to debug my libz problem) is not finding the hdf5 libraries when importing netcdf4, even though the hdf5 package was installed. After reverting back to an older gdal package, the hdf5 and libnetcdf packages were downgraded:
```
The following packages will be DOWNGRADED:

    curl:      7.38.0-0 --> 7.30.0-0 
    hdf5:      1.8.14-0 --> 1.8.9-1  
    libnetcdf: 4.3.2-1  --> 4.2.1.1-1
```
Perhaps there is a conflict between the netcdf4 python package and the libnetcdf package? It would seem that libnetcdf got upgraded without a corresponding upgrade to the netcdf4 package. Perhaps the netcdf4 package is trying to get the libhdf5_hl.so.7 file when it shouldn't be?

WeatherGod:
To be clear, it was the "--with-libz=internal" that enabled it to complete its build. If I did "--with-libz=$PREFIX", I got linker errors.

WeatherGod:
adding the netcdf4 package to *both* the build and run entries in meta.yaml fixed that one particular problem for me, but I have zero clue if that is how one should do it.

gdal/build.sh
asmeurer: On OS X at least it's better to use the system curl.

PR #255, opened 2015-03-01
Build HDF5 C++ library

dfroger:
Some projects, like [ITK](http://www.itk.org/), [use](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/IO/HDF5/include/itkHDF5ImageIO.h#L37) the HDF5 C++ library.

So packaging ITK with conda would require HDF5 to be built with its C++ library.


asmeurer:
@ilanschnell

dfroger:
I close the issue because it is exactly the same as #233, sorry!

PR #257, opened 2015-03-04
imageio v1.2

almarklein:
I was wondering, of all the packages in this list, which one are actually build by Continuum, and how is that decided?

### DIFF
--- a/imageio/meta.yaml
+++ b/imageio/meta.yaml
@@ -1,33 +1,42 @@
 package:
   name: imageio
-  version: 1.0
+  version: 1.2
 
 source:
-  fn: imageio-1.0.tar.gz
-  url: https://pypi.python.org/packages/source/i/imageio/imageio-1.0.tar.gz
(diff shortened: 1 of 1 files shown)


asmeurer:
There's a little more involved getting a package in the official channels, so it's based on popularity of the request, or if a customer requests a package. If you want something included, though, just ping @ilanschnell.

ccordoba12:

almarklein:
Well, I build imageio conda packages myself for py3k (which is me launching 5 different VM's), and then someone asked for one on Python 2.7. He proposed to automate building the packages via Travis + Appveyor, which is certainly an option, but if it would be in the official channels, that would be even easier :)

@ilanschnell Would imageio be popular enough (currently 3k downloads per month on Pypi) to be considered for inclusion?

ilanschnell:
Yes, definitely.  I think at one point we already considered imageio, but had trouble building packages on some platforms.

PR #259, opened 2015-03-26
MNT : add recipe for bzr

tacaswell:


ccordoba12:
I thought bazaar was almost not maintained anymore

tacaswell:
That may be, but I need it on a machine I don't have root on.  There is
still a large number of projects which are hosted in bzr repos.

bzr/meta.yaml
```diff
@@ -0,0 +1,18 @@
...
+  name: bzr
+  version: 2.6.0
+
+source:
+  fn: bzr-2.6.0.tar.gz
+  url: https://launchpad.net/bzr/2.6/2.6.0/+download/bzr-2.6.0.tar.gz
+  md5: 28c86653d0df10d202c6b842deb0ea35
+
+build:
+  number: 1
```
ccordoba12: Since this is a new package, build number must be 0. In other words, you don't need to add a `build/number` section for new packages

tacaswell: I think I have a version on my public channel that has build number 1 (I
think I used zero testing), is it worth worrying about the global build
number?

ccordoba12: It's not that serious, but all new packages starts with build number 0. It's just how things work :-)

ccordoba12:
Thanks for the change @tacaswell. Merging now :-)

PR #260, opened 2015-03-26
Update qt5, pyqt5, and sip. Also fixes qt5 build on unix

jdreaver:
I have done the following to these packages:
* Update qt5 to version 5.4.1
* Update pyqt5 to version 5.4.1
* Update sip to version 4.16.7
* Fix the qt5 build on unix by removing the Windows-only `-process` flag. This was probably just copy-and-pasted on accident.

I have tested the resulting pyqt5 package by creating a simple QtWidgets view and also a simple QtQuick application. Therefore, I am confident that the main modules like QtCore, QtWidgets, QtGui, and QtQuick work.

I made builds of all three of these packages on Arch Linux and uploaded them to binstar (my username is jdreaver there too). Here is a [link to the pyqt5 linux file download page](https://binstar.org/jdreaver/pyqt5/files).

I haven't tested these updates on Windows. I did update the SHA1 sums of the source zip files though. Of course, more people should probably try this recipe out on different Linux distributions and on OSX too :)

### DIFF
--- a/pyqt5/meta.yaml
+++ b/pyqt5/meta.yaml
@@ -1,14 +1,14 @@
 package:
   name: pyqt5
-  version: 5.4
+  version: 5.4.1
 
 source:
-  fn: PyQt-gpl-5.4.zip # [win]
-  url: http://sourceforge.net/projects/pyqt/files/PyQt5/PyQt-5.4/PyQt-gpl-5.4.zip # [win]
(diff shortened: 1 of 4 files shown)


ccordoba12:
Thanks for your work. It's really good to know that the recipes are working on Linux. After Anaconda 2.2 I'll test them on Windows and Mac.

PR #261, opened 2015-04-03
Added linux version of icu

jdreaver:
I simply added icu for linux32 and linux64. My only test was running it with PyQt5, and no errors were reported. Please tell me if something doesn't look right.

### DIFF
--- a/icu/build.sh
+++ b/icu/build.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
-cp -r usr/local/include/ "$PREFIX/include/"
-cp -r usr/local/lib/ "$PREFIX/lib/"
-cp -r usr/local/bin/ "$PREFIX/bin/"
+cd source
+chmod +x runConfigureICU configure install-sh
+./runConfigureICU Linux --prefix="$PREFIX"
+make
+make install
--- a/icu/build.sh
+++ b/icu/build.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+cp -r usr/local/include/ "$PREFIX/include/"
+cp -r usr/local/lib/ "$PREFIX/lib/"
+cp -r usr/local/bin/ "$PREFIX/bin/"
--- a/icu/meta.yaml
+++ b/icu/meta.yaml
@@ -9,12 +9,9 @@ source:
   fn: icu4c-54_1-Win32-msvc10.zip # [win32]
   url: http://download.icu-project.org/files/icu4c/54.1/icu4c-54_1-Win32-msvc10.zip # [win32]
   md5: 2419927dae9f0ac2498b10de657ff6d3 # [win32]
-  fn: icu4c-54_1-RHEL6-x64.tgz # [linux64]
-  url: http://download.icu-project.org/files/icu4c/54.1/icu4c-54_1-RHEL6-x64.tgz # [linux64]
-  md5: 9ac1a7b4f471e83115b42b7ce80cc06c # [linux64]
-  fn: icu4c-54_1-RHEL6-i386.tgz # [linux32]
-  url: http://download.icu-project.org/files/icu4c/54.1/icu4c-54_1-RHEL6-i386.tgz # [linux32]
-  md5: 535555d4e142f6004be8c602c6104298 # [linux32]
+  fn: icu4c-54_1-src.tgz # [linux]
+  url: http://download.icu-project.org/files/icu4c/54.1/icu4c-54_1-src.tgz # [linux]
+  md5: e844caed8f2ca24c088505b0d6271bc0 # [linux]
 
 about:
     home: http://site.icu-project.org/
--- a/icu/meta.yaml
+++ b/icu/meta.yaml
@@ -9,6 +9,12 @@ source:
   fn: icu4c-54_1-Win32-msvc10.zip # [win32]
   url: http://download.icu-project.org/files/icu4c/54.1/icu4c-54_1-Win32-msvc10.zip # [win32]
   md5: 2419927dae9f0ac2498b10de657ff6d3 # [win32]
+  fn: icu4c-54_1-RHEL6-x64.tgz # [linux64]
+  url: http://download.icu-project.org/files/icu4c/54.1/icu4c-54_1-RHEL6-x64.tgz # [linux64]
+  md5: 9ac1a7b4f471e83115b42b7ce80cc06c # [linux64]
+  fn: icu4c-54_1-RHEL6-i386.tgz # [linux32]
+  url: http://download.icu-project.org/files/icu4c/54.1/icu4c-54_1-RHEL6-i386.tgz # [linux32]
+  md5: 535555d4e142f6004be8c602c6104298 # [linux32]
 
 about:
     home: http://site.icu-project.org/


icu/meta.yaml
```diff
@@ -9,6 +9,12 @@ source:
   fn: icu4c-54_1-Win32-msvc10.zip # [win32]
   url: http://download.icu-project.org/files/icu4c/54.1/icu4c-54_1-Win32-msvc10.zip # [win32]
   md5: 2419927dae9f0ac2498b10de657ff6d3 # [win32]
+  fn: icu4c-54_1-RHEL6-x64.tgz # [linux64]
```
ccordoba12: We try to build everything in CentOS 5. Are there sources available for that platform?

icu/build.sh
```diff
@@ -0,0 +1,3 @@
+cp -r usr/local/include/ "$PREFIX/include/"
```
ccordoba12: Shebang to bash is missing here

ccordoba12:
@Juanlu001, providing our own icu would be of better help when building boost, right?

Juanlu001:
Probably so! In general, having to rely in the system packages (that is,
stepping out of conda) defeats the advantages of having a conda package and
causes linking problems on runtime to those that don't install them.

icu/meta.yaml
jdreaver: Unfortunately not :/ [Here is the download page.](http://site.icu-project.org/download/54#TOC-ICU4C-Download)

jdreaver:
Thanks for the quick review! I added the shebang line to `build.sh` and just amended the commit.

I added a Linux `icu` version because it turns out `qt5` depends on it. I am still trying to get `qt5` to work well, and the biggest problem is different versions of dependency libs on different linux distributions. The sane thing to do, in my opinion, is try and get all of the main dependencies in conda.

icu/meta.yaml
Juanlu001: There are binaries for CentOS 5 if you go back to older versions, see for instance http://icu-project.org/download/4.4.html#ICU4C

Juanlu001: By the way, those are precompiled binaries, not sources.

jdreaver: Oh, I didn't notice that. I guess we could modify this later to compile from source. I just wanted to do exactly what the current windows build does, but on Linux (and it seems to work).

ccordoba12:
@jdreaver, what dependencies are missing?

jdreaver:
The following dependencies gave me trouble:
* `pcre`: I tried compiling with `-qt-pcre`, but when I run `ldd` on, for example, `libQt5Widgets.so`, it still points to the system `libpcre`. When I run on a system with the conda `pcre` but maybe an older system `libpcre`, then Qt5 complains that it cannot find `pcre`.
* `icu`: This was fixed by adding `icu` as a dependency in conda.

When I compiled with Arch Linux (which usually has the latest versions of libraries), I got errors on Ubuntu about missing `pcre`, but when I compile with Ubuntu 14.04, I get no errors on Arch Linux. Therefore, the `qt5` package on my binstar page was compiled with Ubuntu, and `icu` was set as a build and runtime dependency for linux. I was going to open a pull request for the `qt5` recipe once I sorted out all of the issues.

Currently, I am getting some sort of OpenGL error on my headless CI box and on Travis for another project (pytest-dev/pytest-qt#35). I can't tell if it is an issue with `xcb`, `xvfb`, or the fact that Qt5 requires OpenGL 2.0 or greater, and a lot of headless servers only have OpenGL 1.1.

Anyway, I am going to test it on my other laptops with the GUI version of Ubuntu to see if all is well. I guess the `qt5` recipe just requires some more work and some more testing!  :smile:

ccordoba12:
`libpng` and `libz` (the same as `zlib`, right?) are part of conda already, so we should only need to add `icu` and `pcre`.

I'll start testing and tweaking our the `qt5` recipes at the end of the month, but I need to add `icu` right now for `boost`.

ccordoba12:
@jdreaver, could you change to use the CentOS 5 binaries, as suggested by @Juanlu001, so I can merge this PR?

jdreaver:
I checked out the link and I didn't see any CentOS binaries. Also, the link @Juanlu001 posted was from a version of `icu` that is 5 years old, and there certainly aren't CentOS binaries now (at least from what I see). 

I was saying that in the future we could compile from source, but for now the RHEL binaries in the package work for me on both Ubuntu 14.04 and Arch Linux.

Please correct me if there are CentOS binaries on the page. Ctrl+F "Cent" comes up empty. All I see are AIX, RHEL, Solaris, Win64, and Win32 binaries :/

Juanlu001:
Actually RHEL and CentOS are equivalent, right? The latter is based on the
former. Surely the RHEL binaries can be used.

jdreaver:
In that case, is this good to merge?

ccordoba12:
I think we should provide a recipe for ICU that actually compiles it on Linux. How hard do you think that would be?

jdreaver:
Done! Compiling from source turned out to be easier than I thought. :smile: Another advantage is there is no difference between Linux32 and Linux64.

I tested it by running a PyQt5 application. I verified that `ldd libQt5Widgets.so` points to the `icu` I just compiled and not the system `icu`, so I believe it works.

ccordoba12:
Great man! Really great job!

jdreaver:
Thanks!

jdreaver:
@ccordoba12 (maybe @asmeurer will have input too), I have a question about `qt5` and CentOS. I put it here because you mentioned CentOS 5 in this thread.

I am trying to build the `qt5` recipe on CentOS 5, and I am having trouble getting recent enough dependencies (take a look at [this article](http://kate-editor.org/2014/12/22/qt-5-4-on-red-hat-enterprise-5/) just to get a taste of all the extra steps required). I'm sure I can eventually figure it out, but I am worried about how reproducible and *general* the process will be. How should we manage all of the extra `qt` dependencies like `libxcb-*`, `libX11`, etc?

1. Should we try to make recipes for all of the `qt5` dependencies in `conda`? There is [quite a large list of them](http://doc.qt.io/qt-5/linux-requirements.html), especially if we are building QtWebKit too. We already do this with `icu` and `freetype`, so we just have to continue with the rest of the dependencies.
2. Should we add an if-statement in `build.sh` for CentOS 5 to install these dependencies? For many libraries, building with an older version will allow Qt5 to work when a newer version is present on the system. For example, I build `qt5` with an old `pcre` version on Ubuntu, and it worked when I had a newer version of `pcre` on my Arch Linux system, but it does not work the other way around. This approach of relying on system packages feels dirty to me.

On the one hand, I think that having the conda `qt5` package being totally standalone is ideal. On the other hand, I don't find it totally unreasonable to require certain packages that might be installed already on an OpenGL-enabled GUI-based Linux distribution. Also, I am not an expert in `xcb` and OpenGL, so I don't know if the `libxcb` and `libX11` libraries are actually portable across distributions and can be simply included in `conda`.

I guess more info on the CentOS build environment you use would help, and any other input would be appreciated. I am leaning towards including all `qt5` deps in conda, personally. I just know it will take a lot of work :smile:

ccordoba12:
@jdreaver, I've already seen the page you mentioned :-) I think we only need to add three new deps for Qt5 (without considering QtWebkit):

1. xcb-proto
2. pthread
3. xcb

The others can be installed using yum (that's what we do right now to compile Qt4 :-)

To compile QtWebkit we'd also need:

1. xrender
2. gstreamer
3. flex(?)

I think gstreamer optional, at least that's the case with Qt4.

jdreaver:
Great! I've added `xcb-proto`, `libxcb`, and `pthread-stubs` to [my fork](https://github.com/jdreaver/conda-recipes) so far. I can add a WIP pull request if you like. 

Do you think it is feasible to use the `-qt-xcb` configure flag instead of adding an `xcb` recipe, or is `libxcb` good enough?

Thanks for the help!

ccordoba12:
I haven't compiled Qt5 before, so I didn't know about `-qt-xcb`. I think that's much better than using our own xcb versions. I think the less packages we need to rely on for Qt5 the better :-)

It seems `-qt-xcb` also provides xrender, pthread, fontconfig and freetype according to

http://doc.qt.io/qt-5/linux-requirements.html#platform-plugin-dependencies

so only gstreamer would be missing!!

By the way, there is also a `-qt-pcre` flag which maybe compiles a built-in version of pcre.

And finally, yes, please open a WIP PR so I can start testing it on my side too.

PR #264, opened 2015-04-08
[qt5] Require icu for Linux build

jdreaver:
Just like we discussed in #261.


ccordoba12:
Let's merge #261 before this one :-)

PR #265, opened 2015-04-09
[WIP] Adding Qt5 Linux dependencies

jdreaver:
As discussed in #261, we want to try and use configure flags like `-qt-pcre` and `-qt-xcb` so we don't have to pull in too many extra dependencies. Note that `-qt-xcb` relies on a lot of `libxcb-*` libraries, so `libxcb` should be a dependency of `qt5`. Also, `libxcb` requires `xcb-proto` and `pthread-stubs`, so I have added those.

The next step, I think, is to add `gstreamer`. 

I tried building on a CentOS 5 vagrant box, but the build fails right after symlinking the `libQtQWidgets.so*` libraries. It could be that more dependencies are missing. I've never used CentOS before, so this could be because I missed a very obvious build package :)

### DIFF
--- a/qt5/meta.yaml
+++ b/qt5/meta.yaml
@@ -18,30 +18,21 @@ requirements:
     - freetype # [unix]
     - jom # [win]
     - python # [win]
-    - gcc # [linux]
-    - libxcb # [linux]
-    - xcb-util # [linux]
-    - xcb-util-image # [linux]
-    - xcb-util-keysyms # [linux]
-    - xcb-util-wm # [linux]
-    - libx11 # [linux]
     - fontconfig # [linux]
     - gstreamer # [linux]
     - bison # [linux]
     - flex # [linux]
     - sqlite # [linux]
     - icu
+    - zlib # [linux]
+    - xz # [linux]
   run:
     - freetype # [unix]
     - fontconfig # [linux]
-    - libxcb # [linux]
-    - xcb-util # [linux]
-    - xcb-util-keysyms # [linux]
-    - xcb-util-image # [linux]
-    - xcb-util-wm # [linux]
-    - libx11 # [linux]
     - gstreamer # [linux]
     - icu
+    - zlib # [linux]
+    - xz # [linux]
 
 
 about:
--- a/qt5/meta.yaml
+++ b/qt5/meta.yaml
@@ -20,6 +20,11 @@ requirements:
     - python # [win]
     - gcc # [linux]
     - libxcb # [linux]
+    - xcb-util # [linux]
+    - xcb-util-image # [linux]
+    - xcb-util-keysyms # [linux]
+    - xcb-util-wm # [linux]
+    - libx11 # [linux]
     - fontconfig # [linux]
     - gstreamer # [linux]
     - bison # [linux]
@@ -30,6 +35,11 @@ requirements:
     - freetype # [unix]
     - fontconfig # [linux]
     - libxcb # [linux]
+    - xcb-util # [linux]
+    - xcb-util-keysyms # [linux]
+    - xcb-util-image # [linux]
+    - xcb-util-wm # [linux]
+    - libx11 # [linux]
     - gstreamer # [linux]
     - icu
 
--- a/qt5/meta.yaml
+++ b/qt5/meta.yaml
@@ -18,11 +18,18 @@ requirements:
     - freetype # [unix]
     - jom # [win]
     - python # [win]
+    - gcc # [linux]
     - libxcb # [linux]
+    - fontconfig # [linux]
+    - gstreamer # [linux]
+    - bison # [linux]
+    - flex # [linux]
     - icu
   run:
     - freetype # [unix]
+    - fontconfig # [linux]
     - libxcb # [linux]
+    - gstreamer # [linux]
     - icu
 
 
--- a/qt5/meta.yaml
+++ b/qt5/meta.yaml
@@ -24,6 +24,7 @@ requirements:
     - gstreamer # [linux]
     - bison # [linux]
     - flex # [linux]
+    - sqlite # [linux]
     - icu
   run:
     - freetype # [unix]
--- a/qt5/meta.yaml
+++ b/qt5/meta.yaml
@@ -18,9 +18,11 @@ requirements:
     - freetype # [unix]
     - jom # [win]
     - python # [win]
+    - libxcb # [linux]
     - icu
   run:
     - freetype # [unix]
+    - libxcb # [linux]
     - icu
 
 
--- a/qt5/meta.yaml
+++ b/qt5/meta.yaml
@@ -18,6 +18,8 @@ requirements:
     - freetype # [unix]
     - jom # [win]
     - python # [win]
+    - libpng # [linux]
+    - jpeg # [linux]
     - fontconfig # [linux]
     - gstreamer # [linux]
     - bison # [linux]
@@ -28,6 +30,8 @@ requirements:
     - xz # [linux]
   run:
     - freetype # [unix]
+    - libpng # [linux]
+    - jpeg # [linux]
     - fontconfig # [linux]
     - gstreamer # [linux]
     - icu
--- a/bison/build.sh
+++ b/bison/build.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+./configure --prefix="$PREFIX"
+make
+make install
--- a/bison/meta.yaml
+++ b/bison/meta.yaml
@@ -1,0 +1,16 @@
+package:
+  name: bison
+  version: 3.0.4
+
+source:
+  fn: bison-3.0.4.tar.gz # [linux]
+  url: http://ftp.gnu.org/gnu/bison/bison-3.0.4.tar.gz # [linux]
+  md5: a586e11cd4aff49c3ff6d3b6a4c9ccf8 # [linux]
+
+about:
+    home: https://www.gnu.org/software/bison/
+    license: GPL3
+
+requirements:
+  build:
+    - m4 # [linux]
--- a/bison/meta.yaml
+++ b/bison/meta.yaml
@@ -14,3 +14,6 @@ about:
 requirements:
   build:
     - m4 # [linux]
+  
+  run:
+    - m4 # [linux]
--- a/damageproto/build.sh
+++ b/damageproto/build.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+./configure --prefix="$PREFIX"
+make
+make install
--- a/damageproto/meta.yaml
+++ b/damageproto/meta.yaml
@@ -1,0 +1,12 @@
+package:
+  name: damageproto
+  version: 1.2.1
+
+source:
+  fn: damageproto-1.2.1.tar.bz2 # [linux]
+  url: http://xorg.freedesktop.org/releases/individual/proto/damageproto-1.2.1.tar.bz2 # [linux]
+  md5: 998e5904764b82642cc63d97b4ba9e95 # [linux]
+
+about:
+    home: http://xorg.freedesktop.org/
+    license: MIT
--- a/dri2proto/build.sh
+++ b/dri2proto/build.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+./configure --prefix="$PREFIX"
+make
+make install
--- a/dri2proto/meta.yaml
+++ b/dri2proto/meta.yaml
@@ -1,0 +1,12 @@
+package:
+  name: dri2proto
+  version: 2.8
+
+source:
+  fn: dri2proto-2.8.tar.gz # [linux]
+  url: http://xorg.freedesktop.org/releases/individual/proto/dri2proto-2.8.tar.gz # [linux]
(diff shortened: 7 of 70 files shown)


jdreaver:
I have a recipe for `gstreamer`, but it requires `bison`, which in turn requires `m4`. I think it is reasonable to install `bison` as a system package, but the CentOS 5 version of `bison` is too old. I can make recipes for `bison` and `m4` too if need be.

When you guys use CentOS to build packages, are you using a more up-to-date yum repo than the one that is bundled with CentOS 5?

jdreaver:
Another note: I had to install `gettext` in CentOS 5 to build `glib`. I also had to build `libffi` myself, because when I used the official Anaconda one, the configure step complained that `libffi` wasn't there.

I went ahead and just added recipes for `flex` and `bison`, since gstreamer depends on them.

jdreaver:
For some reason, libtools was looking for `libffi.la` inside of the nonexistent `lib64` directory, not the `lib` directory. In `gstreamer`'s `build.sh`, I temporarily move it there, but I don't know if this will break `gstreamer`. Just throwing this out there if anyone gets `libffi` or `gstreamer` errors.

ccordoba12:
> When you guys use CentOS to build packages, are you using a more up-to-date yum repo than the one that is bundled with CentOS 5?

Nop. I just added the EPEL repo but only to install git.

So is Qt5 still failing to compile?

ccordoba12:
Another thing: What if you start without building QtWebkit?

jdreaver:
Yeah `qt5` is still failing to compile. I will try without QtWebKit if I encounter a problem there, but I think the problem right now is in `qtbase`, which is just QtGui, QtCore, and QtWidgets. I think it is some sort of OpenGL library error, but no useful error message is printed. I'm still investigating.

I just pushed the latest modifications to the `qt5` recipe so far in case you want to see where I'm at.

qt5/meta.yaml
```diff
@@ -18,9 +18,11 @@ requirements:
     - freetype # [unix]
     - jom # [win]
     - python # [win]
+    - libxcb # [linux]
     - icu
```
ccordoba12: I think you should add here `gcc` as a build dependency. `gcc` is now on the Continuum repos and it's at version 4.8.2.

fontconfig seems to be missing as a build dependency too.

jdreaver:
The `gcc` and `fontconfig` suggestion got me further! Now I am getting a different compilation error. I am piping the build output to a file so I can investigate it better.

jdreaver:
The latest error is kind of tricky. I am getting 4 errors similar to this:
```
In file included from /usr/include/asm-x86_64/byteorder.h:30:0,
                  from /usr/include/asm/byteorder.h:5,
                  from 3rdparty/linux_perf_event_p.h:19,
                  from qbenchmarkperfevents.cpp:53:
/usr/include/linux/byteorder/little_endian.h:43:19: error: ‘__le64’ does not name a type
  static __inline__ __le64 __cpu_to_le64p(const __u64 *p)
```

After some research, I found that [this guy](http://lists.qt-project.org/pipermail/interest/2015-February/015306.html) found the same error, and everyone told him it was CentOS 5's fault. After even more research, I found [this message[(https://gcc.gnu.org/ml/gcc-patches/2014-11/msg01767.html) concerning the same error, but not with Qt5. Basically, there is an `#ifdef` in `/usr/include/linux/types.h` that looks like this:

```
#if defined(__GNUC__) && !defined(__STRICT_ANSI__)
typedef __u64 __bitwise __le64;
typedef __u64 __bitwise __be64;
#endif
```
The way to activate this is to compile with `-std=gnu++11`, but the compilation command that triggers this command is set to `-std=c++0x`.

1. Is there a way to patch and/or override `/usr/include/linux/types.h`?
2. Is there a way to `#define __GNUC__` before compiling so the if-statement passes? Maybe by patching the file we are trying to compile?

I found a tip in [the article](http://kate-editor.org/2014/12/22/qt-5-4-on-red-hat-enterprise-5/) we discussed earlier that says to disable perf events. I'll try this too. However, I hope this doesn't break QTest, as we use that for pytest-qt!

jdreaver:
I can now successfully build the `qt5` package on CentOS 5, and the resulting package works on my Arch Linux box. However, I am getting some OpenGL errors on my headless Ubuntu 14.04 server right now (it ran the old recipe fine), so there is more work to be done.

jdreaver:
Is there a reason the version of `glibc` in the conda recipe is really old, and there is a separate `glibc-2.5.1` recipe? It turns out `sip` requires `glibc`, and I would like to upgrade the `glibc` recipe if I can.

ccordoba12:
> Maybe we could just patch linux_perf_event_p.h

Yes, that's the best alternative because that file is part of the Qt source.

> I am getting some OpenGL errors on my headless Ubuntu 14.04 server right now

Maybe you need to add the new xcb packages as runtime deps of Qt5 (not just as build ones).

> I am also getting an error in QtNetwork

Maybe that's generated because you need to add conda's openssl package as build and runtime dep for Qt5

> It turns out sip requires glibc

Why? Are you unable to build sip on CentOS 5? We shouldn't depend on glibc for that (we're not Gentoo after all ;-)

jdreaver:
I just rebuilt `sip` and it worked flawlessly. I think what happened was I tried to build `pyqt5` with my version of `sip` that was created on Arch Linux, so I got a `glibc` error and I forgot which package I was building :sweat_smile: 

I'm rebuilding all of the deps on CentOS 5, and then I will troubleshoot any `qt5` errors. I am not intimately familiar with the `sip`/`PyQt5` build process, but it is possible that after rebuilding `pyqt5` with the new build of `qt5` some problems will be fixed. Thanks for the suggestions!

By the way, have you guys considered using NixOS for official builds? For each package, you could explicitly state any non-conda dependencies in a `.nix` file, and ensure that they are the only system packages in the build environment. Of course, maybe the ideal situation is that conda contains as many of the build dependencies as possible. I just thought I would throw the thought out there while I had it :smile:

jdreaver:
As expected, rebuilding `sip` and `pyqt5` removed the `QtNetwork` error about a missing symbol. 

Now I just have the same OpenGL GLX/EGL error on all of my platforms, as shown in [this thread](https://forum.qt.io/topic/47359/qxcbintegration-cannot-create-platform-opengl-context-neither-glx-nor-egl-are-enabled). I already added `libxcb` as a runtime dependency to `qt5`. I think I just need another system OpenGL package to be installed, or set another configure flag. I'll take a look.

jdreaver:
I think I've narrowed down the OpenGL error: we need `libX11-xcb`, which is build by compiling `libX11` with `libxcb` installed. CentOS 6 has a `libX11-xcb` package, but unfortunately CentOS 5 does not. I suppose we could add a conda `X11` package that relies on `libxcb`, since we already have a `libxcb` package.

jdreaver:
I added some `xcb` and `X11` dependencies instead of using the `-qt-xcb` option. I now get a confirmation that `GLX` support is enabled after `./configure`. However, now I am getting errors about undefined `GLX_*` variables. After further investigation, I think the version of `mesa` in CentOS 5 is too old, and doesn't define these symbols in its `glx_*.h` headers.

I am starting to think we should just try to compile `qt5` on CentOS 6 if possible. As you can see in the `qt5` recipe, there are now a ton of dependencies. I can keep adding more dependencies, but it will take time. Or, we can find a way to get some newer mesa libs on CentOS 5.

jdreaver:
I tried to add `mesa` and its dependencies, and I got really far, but now I encounter [this error](https://bugs.gentoo.org/show_bug.cgi?id=452180). Basically, the CentOS 5 kernel is too old to compiler `libdrm` because it leaves `O_CLOEXEC` undeclared. This might be tough to solve on CentOS 5 (unless there is a way to update the kernel).

ccordoba12:
Thanks a lot for your hard work on this!!

We compile things on CentOS 5 to make sure our libraries are compatible with the oldest systems people are using and we are able to handle. But what we really take advantage of is of the quite old glibc version present in CentOS, so we don't get those obnoxious glibc compatibility errors.

So one idea would be this: compile Qt5 on CentOS 6, but making it depend on the glibc recipe present here but using a version as old as the one in CentOS 5.

I don't know if that's possible, but it seems much better than the current approach. What do you think @jdreaver?

jdreaver:
Thanks! Using CentOS 6 with an old glibc could be a solution. 

Another idea I had was to use older versions of some of the dependencies. If a newer kernel causes errors, and `mesa` is older than that kernel, then certainly there must be a version of `mesa` that works with the older kernel.

I'll try to see if I can work around the issues before trying these ideas.

jdreaver:
I tried compiling on CentOS 6, but now I have a new error. Basically, Qt5 is compiling with the system's `libpng.so.12` instead of the conda `libpng.so.15`. I think `libpng` might be a dependency of `mesa` or an X11 lib, which I am currently installing on the system. Is there a way to force Qt5 to use the libpng from conda?

ccordoba12:
I think you need to use the `$LIBRARY_PATH` and `$INCLUDE_PATH` variables.

Please take a look at how we use them in our Qt4 recipe.

ccordoba12:
Another thing: are all the dependencies you added necessary on CentOS 6?

Also, why you haven't added as build deps on `qt5/meta.yml`

jdreaver:
I think you need to use the $LIBRARY_PATH and $INCLUDE_PATH variables.

Will do.

    Another thing: are all the dependencies you added necessary on CentOS 6?

No. I changed the `notes.md` file to reflect this. Basically, you just need to `sudo yum install *mesa*` to get all of the required mesa and X11 dependencies (note the wildcard character `*`, which pulls *all* mesa packages). Once I tried building on CentOS6, I realized we don't need them, hence they are not build dependencies.

I'll get to this again sometime this week probably. Thanks!

ccordoba12:
Great, thanks for the update!

jdreaver:
It looks like adding `-L $LIBRARY_PATH` and `-I $INCLUDE_PATH` worked! The last problem I have is with xcb. The bundled Qt xcb depends on the particular `libudev` version on the build machine. In this case, CentOS 6 has `libudev.so.0` and Arch Linux has `libudev.so.1`, so the build didn't work on Arch Linux. If need be, I can finish the `libudev` recipe, but for now I am currently building *without* `-qt-xcb`, and am relying on the CentOS `xcb` packages. It looks like the Qt4 conda package does this. I'll see how it goes.

I have a question: why does the Qt4 recipe rely on `-qt-zlib` and `-qt-libpng` if those libs are already in conda? Was there some kind of version mismatch?

ccordoba12:
> If need be, I can finish the libudev recipe, but for now I am currently building without -qt-xcb, and am relying on the CentOS xcb packages

I think the best strategy is to rely on the CentOS packages, as we do for Qt4 :-)

> I have a question: why does the Qt4 recipe rely on -qt-zlib and -qt-libpng if those libs are already in conda? Was there some kind of version mismatch?

I'm not really sure. We preferred to use the internal versions at some point and that stick for the following versions.

ccordoba12:
One question: are you compiling Qt5 with our glibc as a build dep?

jdreaver:
Right now I am not using the conda glibc, because I wanted to remove one extra variable.

jdreaver:
I guess `libudev.so.0` vs `libudev.so.1` in Qt5 is an open issue. Here is a list of bug reports that involve `libudev` somehow:
* [QTBUG-45419](https://bugreports.qt.io/browse/QTBUG-45419)
* [QTBUG-42805](https://bugreports.qt.io/browse/QTBUG-42805)
* [QTBUG-34176](https://bugreports.qt.io/browse/QTBUG-34176)

On my Arch Linux machine, I symlinked `libudev.so.0 -> libudev.so.1` like the first bug report suggested, and everything worked, so it is kind of a pain that the version matters. For now, I am adding `-no-libudev` to the build script.

jakirkham:
FWIW, that bison recipe work on Mac also (OS 10.9).

jdreaver:
Great! 

I think we actually won't end up needing a lot of the dependencies I added. I realized I was going down a rabbit hole of adding dependencies into conda, but we instead decided to just use some system packages. As long as the packages used during the build are old enough, the Qt5 build tends to work on newer systems with newer libs with no problem. 

Btw, I am still interested in this pull request. So far this recipe works fine for my needs on Linux. I don't have enough experience with the Windows build system to make the recipe work there. Also, I am leaving for vacation soon for about a month, so I will have to resume work on this when I get back.

ccordoba12:
Ok, no problem :-) Thanks a lot for your work on this @jdreaver!

ccordoba12:
@jdreaver, thanks a ton for your work! With it I could quickly build Qt5 on Linux! :-)

But now that the recipes are updated, is it ok for me to close this PR?

jdreaver:
Awesome! Yeah, go ahead and close this.

Have you tried the recipe on Windows?

ccordoba12:
Not yet. That's work for next year :-)

I really need this to start testing Spyder 3.0 (in beta right now) in Travis. Next year we will move to use Qt5 in Anaconda.

By the way, I'll upload the packages to our Anaconda channel, in case you want to use them:

https://anaconda.org/spyder-ide

PR #266, opened 2015-04-13
Use cairo as a dependency of Graphviz

jakirkham:


ccordoba12:
Simple question: Why cairo is needed by Graphviz?

jakirkham:
Here is a list of Graphviz's dependencies ( http://www.graphviz.org/Download_source.php ). However, you may note that nearly all (looks like all) are optional. This includes ones we have added like `freetype`. So, need is not really the right word here. However, Graphviz's strength is in how it is able to export into different formats for docs, debugging, dependency tracing, etc. As is common with these graphs, they can get quite big and unwieldy so being able to zoom and move around is important.

TL;DR `cairo` is very helpful as it provides lossless formats that are good for interactive viewing of large detailed diagrams, namely, SVG, PDF, PS.

ccordoba12:
Thanks for the detailed explanation! Merging then!

jakirkham:
Hope it wasn't too much. Thanks @ccordoba12.

ccordoba12:
It was perfect!! Don't worry about it ;-)

PR #269, opened 2015-04-15
uwsgi updated from 2.0.2 to 2.0.10

pigmej:
This pull requests updates uwsgi recipe to 2.0.10.
It also adds pcre to build requirements


ilanschnell:
Why did you set the build number to 2?  Any why did you comment out the tests?

pigmej:
Sorry, that was 'dirty' recipe.

Btw, the tests were just "uwsgi --help" which is probably not enough in many cases, but reverted that back. (for example without pcre uwsgi rewrites / redirections are very limited).

Should I squash commits ?

PR #273, opened 2015-04-22
Build cairo without X11 support on Mac

jakirkham:
As one may want cairo, but may not have or want to build X11 to use it.

### DIFF
--- a/cairo/build.sh
+++ b/cairo/build.sh
@@ -2,6 +2,14 @@
 
 export CFLAGS="-I$PREFIX/include -L$PREFIX/lib"
 
+# As of Mac OS 10.8, X11 is no longer included by default ( https://support.apple.com/en-us/HT201341 ).
+# Due to this change, we disable building X11 support for cairo on Mac by default.
+export XWIN_ARGS=""
+if [ `uname` == Darwin ]; then
+   export XWIN_ARGS="--disable-gtk-doc --disable-xlib -disable-xcb --disable-glitz"
+fi
+
+
 ./configure                 \
     --prefix=$PREFIX        \
     --disable-static        \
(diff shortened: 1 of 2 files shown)


ccordoba12:
Why are these changes necessary?

jakirkham:
X11 is no longer included on Mac ( https://support.apple.com/en-us/HT201341 ). This happened as of 10.8.

We don't have the recipes to build it here. Even if we did, it is a bit of a lengthy build. If it is not installed, cairo doesn't seem to be smart enough to skip building the support and just fails instead.

One could install XQuartz ( http://xquartz.macosforge.org/landing/ ). However, it seems risky to have it link to something outside conda that does not come with the system out of the box (i.e. we can always expect to find it).

If we really wanted to use a GUI with cairo a better choice might be Qt as it is included conda and cairo supports this as of 1.9.4 ( http://cairographics.org/news/cairo-1.9.4/ ), which came out 5.5 years ago. This would also make it easier to extended to Windows in the future.

If you have thoughts on this, I'd be interested in hearing them.

ccordoba12:
Ok, I understand your argument and totally agree with it.

But I think on Linux people would expect a cairo lib built with X11 support, so you need to split the script on Linux and Mac parts. Just take at our qt recipe to see how to do it :-)

jakirkham:
Ok, that sounds perfectly reasonable. Let me work on some changes and get back to you

jakirkham:
Sorry for the long turn around on this @ccordoba12. I made some changes that I hope you will find acceptable to check for Mac before disabling X11. I built this successfully on Mac OS 10.9 and CentOS 6.3. I noted that on Mac it did not attempt to link against X11 and that on CentOS it did.

ccordoba12:
Thanks for the extra effort!! Merging now ;-)

jakirkham:
No problem. Thanks.

PR #277, opened 2015-04-29
add pip to rodeo runtime deps

mattexx:
this is needed to build for python2 in ubuntu lucid


ccordoba12:

PR #279, opened 2015-04-29
gcc-4.8: Fix Linux portability problems

stuarteberg:
**Edit:** This thread is getting long, so I'll try to keep an up-to-date summary at the top here.

**\<tl;dr\>**Without this patch, the `gcc` package is worthless on nearly all Linux distros.  But with this patch, it works on all non-ancient Linux distros.**\</tl;dr\>**

**\<SUMMARY\>**

Conda's `gcc` package was built on CentOS 5.11, and only works on that distro.  However, we can make it work on more distros by adding a `post-link.sh` file.

In this PR, `post-link.sh` does three important things:

1. Remove the headers generated by gcc's "fixincludes" build step.  They were made specifically for CentOS 5.11, and don't work elsewhere.  If your target OS already has ANSI-compliant libc headers, you don't need the extra "fixed" headers provided by gcc.  (On newer Linux distros, this seems to be the case.)
2. Help gcc locate the system libc headers by adding an `-I` flag to the default gcc flags (via a `specs` file).
3.  Help gcc locate certain system libc object files (e.g. `crt1.o`) by creating some symlinks to them.

Also, `post-link.sh` has the following features:

I. If you are installing onto OSX, do nothing (`post-link.sh` is a no-op).
II. If you are installing onto the **same OS** that was used to create the `gcc` package, do nothing. (Here, the definition of "same OS" means: compare the output of `cat /etc/*-release` at build-time and install-time.)
III. In interest of "failing early" if you're trying this `gcc` package on a system that just can't use it, the `post-link.sh` script compiles a "Hello World" program at install time.  If something is broken, running `conda install gcc` will simply fail.

*Without* this PR, the `gcc` package is basically non-portable across Linux distros.

It's true: even *with* this PR, we still can't support *every* Linux distro under the sun, but at least now we can support a lot more.  And thanks to item `II` above, you can always just build the `gcc` recipe yourself (in which case `post-link.sh` becomes a no-op).  This means the Anaconda build server won't be affected by these changes.

(Sadly, if your Linux distro is so old that it has "bad" libc headers, even this PR can't help you.  In that case your only option is to build gcc yourself, or obtain it through other means.)

Good news! You can try this PR on own system without needing to build gcc yourself:

    conda install -c stuarteberg gcc=4.8.4.99

That package is known to work on the following distros:
- CentOS 5.11
- CentOS 6.3
- CentOS 6.6
- Fedora 20
- Ubuntu 12.04
- Ubuntu 14.04
- Archlinux

**\</SUMMARY\>**

----
This PR attempts to fix the issue described on the mailing list here:
https://groups.google.com/a/continuum.io/d/msg/conda/HwUazgD-hJ0/aofO0vD-MhcJ

To summarize:
- The gcc build scripts run a step called "fixincludes" which generates custom versions of several system header files.
- Those files are compatible ONLY with the system gcc was built on, so including them in the gcc package makes the gcc binary non-portable to other systems.
- Apparently, many Linux distros simply remove these "fixed" headers as a workaround for this problem.  (See http://ewontfix.com/12).  This PR does the same.
- As long as the target system has ANSI-compatible headers (which don't use C language syntax that gcc doesn't support yet), it should be okay to omit the "fixed" versions.

----
**Update:** As described below, now this PR also address two other portability issues: locating system headers and locating system libc code on various distros.


asmeurer:
Should this also run on OS X? If not, add an if statement to the script.

asmeurer:
What is the best way to test this?

stuarteberg:
>Should this also run on OS X? If not, add an if statement to the script.

After some testing on Mac OS 10.10, apparently not.  (I'll update the patch.)  The only "fixed" header is stdint.h, and it doesn't seem to cause compilation errors on the programs I've tried so far.

On a different point, the conda gcc package doesn't work out-of-the-box on Mac OS X.  (It doesn't know where to find the system headers.)  I'll open a separate issue for that later.

>What is the best way to test this?

Well, ideally, we'd just try compiling a big stack (all of the default Anaconda packages?) on multiple Linux distributions.  If nothing seems to break, I guess we're good to go.

stuarteberg:
>Should this also run on OS X? If not, add an if statement to the script.

Done.

BTW, I was wrong about this Mac issue I mentioned above:

>On a different point, the conda gcc package doesn't work out-of-the-box on Mac OS X. (It doesn't know where to find the system headers.) I'll open a separate issue for that later.

The Mac I borrowed for this testing had XCode installed, but didn't have the "xcode command-line developer tools".  Running `xcode-select --install` fixed my build issues

I suppose it would be possible to help boneheads like me by running a post-install script that verifies the command-line tools are installed, but it's probably not necessary -- most devs have them installed already. Also, verifying that they are installed is [not simple](http://stackoverflow.com/q/15371925/162094).

asmeurer:
It was not my intention for gcc to depend on the developer tools. Requiring them (almost) defeats the purpose. What error did you get? According to `conda inspect linkages gcc`, gcc linkages against the system packages `libSystem.B.dylib`, `libgcc_s.1.dylib` and `libiconv.2.dylib` (all in `/usr/lib/`). I suppose `libgcc_s.1.dylib` is the issue. I didn't realize that it would not be installed unless you have the developer tools installed. Or was it another issue?

stuarteberg:
Just to clarify, I'm not reporting a problem with using the `gcc` or `libgcc` package as a **runtime** dependency for pre-built binaries.  As far as I know, they work just fine for that purpose.  I'm interested in using the `gcc` package as my main compiler, for building my own packages from source.

The problem I ran into is that gcc doesn't know where to find the libc headers (e.g. `wchar.h`, `math.h`, etc.).  I think the headers aren't even installed by default (most users would never need them).  If you install XCode, then you'll get a copy, but they're buried inside the XCode package in an SDK directory:

```
$ export SDK_DIR=`xcrun --show-sdk-path`
$ echo $SDK_DIR
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.10.sdk
$ ls $SDK_DIR/usr/include/std*.h
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.10.sdk/usr/include/stdbool.h
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.10.sdk/usr/include/stddef.h
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.10.sdk/usr/include/stdint.h
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.10.sdk/usr/include/stdio.h
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.10.sdk/usr/include/stdlib.h
```

gcc doesn't know about this directory.  But if you install the "Xcode command-line tools", then all the standard headers are copied into `/usr/include`, which gcc DOES know about.

```
$ xcode-select --install
$ ls /usr/include/std*.h
/usr/include/stdbool.h	/usr/include/stddef.h	/usr/include/stdint.h	/usr/include/stdio.h	/usr/include/stdlib.h
```

(Actually, the headers were only part of the problem.  If I provided the headers via `-I/$SDK_DIR/usr/include`, other things, like `crt1.o` were also missing.  Installing the command-line tools fixed that, too.)

I don't think it's unreasonable to expect users who want to compile their own packages to have XCode and the command-line tools already installed.

asmeurer:
> I don't think it's unreasonable to expect users who want to compile their own packages to have XCode and the command-line tools already installed.

It would be nice to avoid it if possible, since installing XCode requires root permissions, but I don't plan to dedicate a ton of effort to work around it myself.

asmeurer:
> Well, ideally, we'd just try compiling a big stack (all of the default Anaconda packages?) on multiple Linux distributions. If nothing seems to break, I guess we're good to go.

I was hoping for something a little more directed, i.e., something that is broken and easily reproducible with the current package.

stuarteberg:
As I mentioned before, this PR works on Fedora 20.  But when I tested on Ubuntu 12.04, I discovered yet more portability problems:

- Ubuntu keeps the std C headers in `/usr/include/x86_64-linux-gnu`, but the conda-built gcc doesn't know to look there.  Therefore, "hello world" can't compile due to missing headers.

```
$ gcc hello.c
In file included from /usr/include/stdio.h:28:0,
                 from hello.c:1:
/home/vagrant/miniconda/envs/delme/lib/gcc/x86_64-unknown-linux-gnu/4.8.2/include-fixed/features.h:338:25: fatal error: sys/cdefs.h: No such file or directory
 #  include <sys/cdefs.h>
                         ^
compilation terminated.
```

Of course, one workaround is to pass explicit `-I` option:

    gcc -I/usr/include/x86_64-linux-gnu hello.c

... but then we hit the next problem:

- The conda gcc binary can't find the C runtime object files `crt1.0`, `crti.o`, and `crtn.o`, resulting in a link error.

```
$ gcc -I/usr/include/x86_64-linux-gnu hello.c
/usr/bin/ld: cannot find crt1.o: No such file or directory
/usr/bin/ld: cannot find crti.o: No such file or directory
collect2: error: ld returned 1 exit status
```

This, too, can be worked around, this time with the `-B` option:

    $ gcc -I/usr/include/x86_64-linux-gnu -B/usr/lib/x86_64-linux-gnu/ hello.c

Instead of requiring users to pass these flags to gcc every time they use it, we can augment the gcc binary ass follows:

- Fix the first issue by adding the `-I` option to the default set via a gcc `specs` file.
- Fix the second issue with symlinks to `crt1.o`, etc. inside `$PREFIX/lib/gcc/...`, which is always searched. (As suggested in Homebrew/linuxbrew#200.)

I've just added a `post-link.sh` script to this PR, which implements both of the above fixes. (I know post-link scripts are to be avoided when possible, but I think it's the right thing to do here.  I guess we shouldn't be surprised that the gcc package is special in some ways.)

gcc-4.8/post-link.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/sh
+
+if [ "$(uname)" != "Darwin" ]; then
```
asmeurer: Hmm, conda-build really ought to have a way to not even include the file on OS X.

asmeurer:
If you have to have a post-link, fine. Just make sure it is very reliable (if it fails, the install will be broken), and very portable.

stuarteberg:
Regarding testing:

>I was hoping for something a little more directed, i.e., something that is broken and easily reproducible with the current package.

Since this issue is about fixing portability issues, the only way to demonstrate the problem(s) is to try using the package on different OSes.

Just to reiterate the above discussion, there are three known portability issues addressed in this PR:

1. The bogus fixincludes step
2. Can't find C library headers
3. Can't find C library runtime object files

Ubuntu 12.04 appears to be affected by all three issues.  To demonstrate issues 2 and 3, try compiling this C program on Ubuntu 12.04 using conda's gcc binary:

    $ cat > hello.c
    #include <stdio.h>
    int main(){}

    $ gcc hello.c

    $ OR:
    $ gcc -I/usr/include/x86_64-linux-gnu hello.c

To demonstrate issue 1, try compiling this C++ program on Ubuntu 12.04:

    $ cat > hello.cpp
    #include <iostream>
    int main(){}

    $ g++ -I/usr/include/x86_64-linux-gnu -B/usr/lib/x86_64-linux-gnu/ hello.cpp

I don't know which other Linux distributions are affected by these issues, but I happen to know that Fedora 20 is affected by issue 1, and @jakirkham's [issue referenced above](https://github.com/conda/conda-build/issues/384) suggests that CentOS 6.3 is also affected.

stuarteberg:
BTW, my binstar package has a pre-packaged version of these changes you can try out on the VM of your choice:

    conda install -c stuarteberg gcc-no-fixincludes

(Disclosure: I didn't actually build it myself.  I just copied your gcc package, removed the `include-fixed` headers, and added the `post-link.sh` file.)

gcc-4.8/post-link.sh
```diff
@@ -0,0 +1,48 @@
...
+
+    # Locate the system's C-runtime object files and link them into the gcc 
+    #  build so they are automatically on the gcc search path.
+    # (The location of these files varies from one system to the next.)
+    C_RUNTIME_OBJ_FILES="crt1.o crti.o crtn.o"
+    SYSTEM_GCC=/usr/bin/gcc
+    
+    DUMMY_PROGRAM="int main(){return 0;}"
+    GCC_CMDS=`echo "$DUMMY_PROGRAM" | $SYSTEM_GCC -v -xc - 2>&1`
```
stuarteberg: There may be a simpler way for me to get the locations of `crt1.o`, etc.  I'll revise this patch on Monday.

jakirkham:
> I don't know which other Linux distributions are affected by these issues, but I happen to know that Fedora 20 is affected by issue 1, and @jakirkham's [issue referenced above]( https://github.com/conda/conda-build/issues/384 ) suggests that CentOS 6.3 is also affected.

@stuarteberg So, I tried your binary ( https://binstar.org/stuarteberg/gcc-no-fixincludes ) and it did indeed resolve my build issues regarding NumPy on CentOS 6.3. It also worked on SciPy.

stuarteberg:
I've added a test step in two places: `run_test.sh` and in `post-link.sh`.  The `post-link.sh` step seems warranted because these portability issues can only be detected on the target system, not the build system.

It's worth noting that the addition of this new test in `post-link.sh` means that `gcc` should never be uses as a `run` dependency.  The `libgcc` package should be used instead.  (BTW, Now would be a good time to move @asmeurer's `libgcc` package from binstar into conda's `defaults`...)

In other news, I've made some minor simplifications to `post-link.sh` and also added a `pre-unlink.sh` script.  I think this PR is basically complete, so I'm hoping for comments.

asmeurer:
You should bump the build number in the meta.yaml

jakirkham:
@stuarteberg What OS's you tested these pre and post link steps?

stuarteberg:
> What OS's you tested these pre and post link steps?

Fedora 20, Ubuntu 12.04, ScientificLinux 6.3, and Mac OS 10.10 (for which I pushed some minor compatibility fixes just now).

> You should bump the build number in the meta.yaml

OK, done. I bumped it from 1 to 3 because somehow the version in conda already has a build number of 2.

stuarteberg:
>>What OS's you tested these pre and post link steps?

>Fedora 20, Ubuntu 12.04, ScientificLinux 6.3, and Mac OS 10.10

Apparently I didn't test thoroughly enough.  The `post-link.sh` test was using the system's `gcc` executable, not the conda-installed `gcc` (I didn't realize the `PATH` variable isn't updated until AFTER `post-link.sh` is executed...).  Anyway, I pushed a fix to the `post-link.sh`, and updated my test package on binstar: `stuarteberg/gcc-no-fixincludes` (build number: 3).

stuarteberg:
This thread was getting long, so I updated the description with a fresh summary of the changes here.  See above.

connie:
This looks promising, and I was having trouble with
```
In file included from /home/connie/anaconda/envs/_build/include/python2.7/Python.h:38:0,
                 from pydas/dassl.c:16:
/usr/include/string.h:548:5: error: unknown type name '__locale_t'
     __locale_t __loc)
     ^
/usr/include/string.h:552:18: error: unknown type name '__locale_t'
      size_t __n, __locale_t __loc)
                  ^
```
which led us to here, but for me, on Ubuntu  12.04, `conda install -c stuarteberg gcc=4.8.2.99` does not work...

```
$ conda install -c stuarteberg gcc=4.8.2.99
Solving package specifications: .................
Warning: 2 possible package resolutions (only showing differing packages):
	[u'gcc-4.8.2.99-linux_portable.tar.bz2'],
	[u'gcc-4.8.2.99-0.tar.bz2'],

Package plan for installation in environment /home/connie/anaconda:

The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    gcc-4.8.2.99               |   linux_portable       251.3 MB

The following NEW packages will be INSTALLED:

    cloog: 0.18.0-0               
    gcc:   4.8.2.99-linux_portable
    gmp:   5.1.2-2                
    isl:   0.12.2-0               
    mpc:   1.0.1-0                
    mpfr:  3.1.2-0                

Proceed ([y]/n)? y

Fetching packages ...
gcc-4.8.2.99-l 100% |################################| Time: 0:00:07  35.03 MB/s
Extracting packages ...
[      COMPLETE      ]|###################################################| 100%
Linking packages ...
/usr/bin/ld: cannot find crt1.o: No such file or directory
/usr/bin/ld: cannot find crti.o: No such file or directory
collect2: error: ld returned 1 exit status
Installation failed: gcc is not able to compile a simple 'Hello, World' program.
Error: Error: post-link failed for: gcc-4.8.2.99-linux_portable##         |  83%


$ conda info
Current conda install:

             platform : linux-64
        conda version : 3.12.0
  conda-build version : 1.12.1
       python version : 2.7.9.final.0
     requests version : 2.7.0
     root environment : /home/connie/anaconda  (writable)
  default environment : /home/connie/anaconda
     envs directories : /home/connie/anaconda/envs
        package cache : /home/connie/anaconda/pkgs
         channel URLs : https://repo.continuum.io/pkgs/free/linux-64/
                        https://repo.continuum.io/pkgs/free/noarch/
                        https://repo.continuum.io/pkgs/pro/linux-64/
                        https://repo.continuum.io/pkgs/pro/noarch/
          config file : None
    is foreign system : False
```

asmeurer:
@stuarteberg you might want to bump the build number on your linux_portable package to avoid the "Warning: 2 possible package resolutions" (fortunately it looks like conda picks the right one, but it's pretty much random).

stuarteberg:
>you might want to bump the build number on your linux_portable package

Good call.  Done.

@connie, Thanks for the bug report.  I've just pushed a fix for it, and also uploaded a new version of the package on binstar.  To make sure you override the old version you downloaded, use the `--force` option.

    conda install -c stuarteberg --force gcc=4.8.2.99

Does that work?

asmeurer:
You should also bump the build number when you push new versions, so that people don't have to use `--force` to update them.

stuarteberg:
Oh, I guess I'm confused about when `--force` is needed.  The package I uploaded does have a new build number.  But on my machine, I think conda didn't want to download the new version without `--force`.  (It just tried to use the cached version, which had the old build number.)

connie:
Thanks for the fix.  I believe it compiled gcc-4.8 correctly. Unfortunately, it did not solve my original string.h header problem.  =(

jakirkham:
@connie, interesting, I found that it had solved this problem for me previously, but haven't had a chance to retest. Also, it was CentOS 6.3. What were you building?

@stuarteberg, it looks like something weird is happening in your version string. It is choosing between build number `linux_portable` and `0`. I suspect this is the cause of the problem.

stuarteberg:
@connie, I've seen that error, too, but I was pretty sure this PR fixes it.  I am using a fresh Ubuntu 12.04 install, and I just verified that my `/usr/include/string.h` is the same as yours:

```
$ head -n548 /usr/include/string.h | tail -n1
			 __locale_t __loc)
```

I'm able to compile this test program:

```
$ cat /tmp/test.c 
#include <Python.h>
#include <string.h>

int main()
{
  printf("Hello, world\n");
  return 0;
}
```

    $ gcc -I ${PREFIX}/include/python2.7 /tmp/test.c
    $ ./a.out
    Hello, world
I don't have any ideas here, other than to double check a few things:
- Did you install `gcc-4.8.2.99` into a fresh environment?
- Did you activate the right environment before attempting to compile?
- Are you able to compile the above test program?

stuarteberg:
Or, better yet, try these commands, and tell me what the last one outputs:

```
$ conda create -y -n testgcc -c stuarteberg --force gcc=4.8.2.99
$ source activate testgcc
$ find $(dirname $(which gcc))/../lib/gcc/*/*/include-fixed
```

In my VM, the last command produces the following output.  If you have more header files than this, something went wrong and I'd like to figure out what it was.

```
$ find $(dirname $(which gcc))/../lib/gcc/*/*/include-fixed
/miniconda/envs/testgcc/bin/../lib/gcc/x86_64-unknown-linux-gnu/4.8.2/include-fixed
/miniconda/envs/testgcc/bin/../lib/gcc/x86_64-unknown-linux-gnu/4.8.2/include-fixed/X11
/miniconda/envs/testgcc/bin/../lib/gcc/x86_64-unknown-linux-gnu/4.8.2/include-fixed/linux
/miniconda/envs/testgcc/bin/../lib/gcc/x86_64-unknown-linux-gnu/4.8.2/include-fixed/README
/miniconda/envs/testgcc/bin/../lib/gcc/x86_64-unknown-linux-gnu/4.8.2/include-fixed/sys
/miniconda/envs/testgcc/bin/../lib/gcc/x86_64-unknown-linux-gnu/4.8.2/include-fixed/bits
/miniconda/envs/testgcc/bin/../lib/gcc/x86_64-unknown-linux-gnu/4.8.2/include-fixed/limits.h
/miniconda/envs/testgcc/bin/../lib/gcc/x86_64-unknown-linux-gnu/4.8.2/include-fixed/syslimits.h
/miniconda/envs/testgcc/bin/../lib/gcc/x86_64-unknown-linux-gnu/4.8.2/include-fixed/sound
```

connie:
I have just tested the hello world program and gcc compiles it properly.  The program executes and works correctly.   When I type 
    
    which gcc

it points to the correct place and correct version (4.8.2)

    /home/connie/anaconda/bin/gcc

connie:
Thanks everyone.  It finally worked.  Turns out that even though I could compile gcc correctly I had to add the stuarteberg channel to my `.condarc` in order for conda to point to the correct version of gcc.  Otherwise it was pointing to the old non-working version.

stuarteberg:
Thanks for the update, @connie.  BTW, from the original error you posted, it looks like you're working with `pydas`.  @rwest recently wrote a recipe (conda/conda-recipes#303) that works for on Mac and Linux (well, with `cython=0.21`).  Is that the one you're using?  I assume you guys must be working together, otherwise that's a strange coincidence.

I suppose you don't need it any more, but a Linux version of the `pydas` package binary can be found on my binstar channel.

jakirkham:
@stuarteberg, I have retest with your latest version in `binstar` and it seems to work fine.

jakirkham:
@asmeurer, if you don't mind me asking, what is still required for this PR to be merged?

asmeurer:
I haven't had a chance to test this. Let me give it a run and see if it works for me.

stuarteberg:
FWIW, I don't anticipate any further changes to this PR.  Scroll back to the [top of this thread](https://github.com/conda/conda-recipes/pull/279#issue-71995325) for an up-to-date summary of what this PR does.

This will allow other Linux distros to use the `gcc` package, while having no effect whatsoever on the Anaconda build. (On the Anaconda build VM, this should all be a big no-op.  It is also a no-op on OSX.)

I've been using the modified `gcc` package from this PR to successfully build a [reasonably large stack][1] of C/C++ packages on Fedora 20 and CentOS 5.11.  (On Mac, these changes are a no-op.)  A few other users have also been able to use it for their own packages (see thread above).

[1]: https://github.com/ilastik/ilastik-build-conda

asmeurer:
No changes are needed for the libgcc package, right?

stuarteberg:
>No changes are needed for the libgcc package, right?

Right.  But while you're at it, it would be nice if you could upload your `libgcc` package to the conda `defaults` channel, so users don't have to use `-c asmeurer`

orodbhen:
@stuarteberg I sucessfully created a conda package for simple-screen-reader using your gcc package, on both Archlinux and CentOS 6.6. They both compile and install without errors. However, when I try to run the program on CentOS, It's unable to to find the correct version of libstdc++. For some reason it was never linked into the lib directory of the anaconda installation. It is, however, present in pkgs/gcc-4.8.2.99-linux_portable/lib. I was able to get it to run by creating the links manually, though. Any idea why it failed to link? I didn't have this problem in Archlinux.

jakirkham:
@orodbhen, please add `libgcc` as a run time dependency to your builds. If that doesn't work, please let us know.

orodbhen:
@jakirkham Yeah, I just realized that I didn't have it as a runtime dep. I installed gcc-4.8.2.99 from @stuarteberg binstar repo, and it runs fine now. libgcc should work as well, since they provide some redundant libs.

So, yes, I can confirm the fix from @stuarteberg works on both Archlinux and CentOS 6.6.

jakirkham:
@orodbhen, there shouldn't be any need to install the `gcc` compilers in the runtime environment. It's kind of a large package and will bulk up anything you are distributing. Making `libgcc` a dependency should provide just enough to make sure anything built continues to function, but will avoid the unneeded compilers and includes.

orodbhen:
I switched the run dep to libgcc on CentOS 6. Still working great. In addition to simple-screen-reader, I also used it to build and package ffmpeg and its various dependencies.

stuarteberg:
@asmeurer: Did you ever get a chance to evaluate this PR? Ideally, I'm hoping for the following:

1. Accept this PR
2. Deploy a fresh build of the `gcc-4.8` package to the `defaults` channel, so everybody can use it out-of-the-box

If there's something I can do to help, let me know.  If you need a refresher on exactly what this PR does, see the up-to-date summary at the [top of this page](https://github.com/conda/conda-recipes/pull/279#issue-71995325).

mjuric:
Hi,
    I'm not sure if you want to handle this corner case, but on a fresh CentOS 6 machine w/o `gcc` installed, the build has problems because it can't run `gcc -print-file-names`:

```bash
The following NEW packages will be INSTALLED:

    gcc: 4.8.2.99-linux_portable

Proceed ([y]/n)? y

Linking packages ...
/home/mjuric/conda-lsst/miniconda/bin/.gcc-post-link.sh: line 41: /usr/bin/gcc: No such file or directory
ln: creating symbolic link `./4.8.2': File exists
/home/mjuric/conda-lsst/miniconda/bin/.gcc-post-link.sh: line 41: /usr/bin/gcc: No such file or directory
ln: creating symbolic link `./4.8.2': File exists
/home/mjuric/conda-lsst/miniconda/bin/.gcc-post-link.sh: line 41: /usr/bin/gcc: No such file or directory
ln: creating symbolic link `./4.8.2': File exists
/home/mjuric/conda-lsst/miniconda/bin/.gcc-post-link.sh: line 41: /usr/bin/gcc: No such file or directory
ln: creating symbolic link `./4.8.2': File exists
/home/mjuric/conda-lsst/miniconda/bin/.gcc-post-link.sh: line 41: /usr/bin/gcc: No such file or directory
ln: creating symbolic link `./4.8.2': File exists
/home/mjuric/conda-lsst/miniconda/bin/.gcc-post-link.sh: line 41: /usr/bin/gcc: No such file or directory
ln: creating symbolic link `./4.8.2': File exists
[      COMPLETE      ]|##################################################################################################################################################################| 100%
[mjuric@centos6 conda-lsst]$ gcc
```

stuarteberg:
>on a fresh CentOS 6 machine w/o gcc installed, the build has problems because it can't run `gcc -print-file-names`

Yeah, my assumption was that users who are interested in using conda's `gcc` package are probably software developers, and therefore already have gcc installed on their system.  (Most conda recipes also require other dev tools to be pre-installed, like cmake, git, hg, etc.)

If I can't assume that `/usr/bin/gcc` exists, then I need some other method of locating the system's `crt0.o`, `crt1.o`, etc. Do you know of a distro-independent way of locating those files?

mjuric:
No, I'm afraid I don't. Assuming `gcc` is there is reasonable, but it would be good to gracefully abort the install if gcc isn't there (the current script continued after the errors, and created some symlinks in my $PWD).

Otherwise, this generally seems to work great! Really good work! Thanks!

mjuric:
Btw., I've found that the GCC binary you're distributing triggers odd `glibc` asserts in [a codebase](http://dm.lsst.org) I'm trying to build to distribute with conda:
```bash
*** glibc detected *** python: double free or corruption (!prev): 0x0000000002c
d15e0 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x75e66)[0x7f04784c7e66]
/lib64/libc.so.6(+0x789b3)[0x7f04784ca9b3]
/home/mjuric/conda-lsst/miniconda/envs/_build/opt/lsst/afw/lib/libafw.so(_ZN5boost6detail17sp_counted_impl_pISt6vectorIdSaIdEEE7disposeEv+0x17)[0x7f046db108a7]
/home/mjuric/conda-lsst/miniconda/envs/_build/opt/lsst/afw/lib/libafw.so(_ZN4lsst3afw4math10Statistics12doStatisticsINS1_13ImageImposterIdEENS1_12MaskImposterItEENS6_IfEES8_EEvRKT_
RKT0_RKT1_RKT2_iRKNS1_17StatisticsControlE+0x573)[0x7f046db225c3]
/home/mjuric/conda-lsst/miniconda/envs/_build/opt/lsst/afw/lib/libafw.so(_ZN4lsst3afw4math10StatisticsC2INS1_13ImageImposterIdEENS1_12MaskImposterItEENS6_IfEEEERKT_RKT0_RKT1_iRKNS1
_17StatisticsControlE+0xa3)[0x7f046db24183]
/home/mjuric/conda-lsst/miniconda/envs/_build/opt/lsst/afw/python/lsst/afw/math/_mathLib.so(+0x394f80)[0x7f0466afdf80]
/home/mjuric/conda-lsst/miniconda/envs/_build/bin/../lib/libpython2.7.so.1.0(PyEval_EvalFrameEx+0x5b0f)[0x7f047918637f]
/home/mjuric/conda-lsst/miniconda/envs/_build/bin/../lib/libpython2.7.so.1.0(PyEval_EvalCodeEx+0x88e)[0x7f0479187dbe]
/home/mjuric/conda-lsst/miniconda/envs/_build/bin/../lib/libpython2.7.so.1.0(PyEval_EvalFrameEx+0x5b8a)[0x7f04791863fa]
....
```
While it's quite possible there's a nasty bug in there that we've yet to find (it's millions lines of code), the same problem doesn't show up with RHEL 6/7+ Ubuntu system-provided compilers, or when I build `gcc-4.8.5` myself.

I wonder if there's something different between 4.8.2 and 4.8.5, or if this could have something to do with your compiler being built on CentOS 5 (I built mine on CentOS 6). Would it be possible to bump the minor `gcc` version in your package to .5?

stuarteberg:
@mjuric - Interesting. I noticed errors like that in my own project, also coming from python.  My, uh, "solution" was to set `MALLOC_CHECK_=0` in my environment before running my code.  Not exactly a bullet-proof strategy, to say the least.

If you have the time, I'd be curious if you can try the following:

- When using your own compiler (not the one from this PR), try setting `export MALLOC_CHECK_=1` in your environment.  Now do the errors appear?
- When using my version of gcc, try compiling with `-fno-builtin-malloc`.  Do the errors go away?

If I get time, one thing I may try is re-compiling gcc with `--enable-checking=release` instead of leaving it as the default.  The [gcc docs][1] say that the `--enable-checking` setting only applies to checks within gcc itself, but it isn't clear to me if that might also change the behavior of its built-in `malloc()`, which would affect its compiled output.

In general, I guess it might be nice if we had a simple test case that exhibits the errors... if you know of one, let me know.

stuarteberg:
Oops, I forgot to mention:

>it would be good to gracefully abort the install if gcc isn't there

Good point.  I just pushed a change to this PR that does just that.  (I haven't yet updated my gcc package on binstar, though.)

mjuric:
@stuarteberg -- the `MALLOC_CHECK_=1` experiment didn't reveal anything (no problems reported when built with a different compiler).

I did go and rebuild the compiler with the recipe from your branch. I think you must've recently rebased it to the latest master -- since @asmeurer bumped the `gcc` version up to 4.8.4, that's what was built (not the 4.8.2 you have on your channel). My build (built on CentOS 5.10, 64bit) is at:
```
conda install -c mjuric gcc
```
(right now the build number is 4).

Bottom line: the problem does not appear with 4.8.4 any more. I suspect we were hitting a real compiler bug in vanilla 4.8.2 (e.g., there were [some nasty bugs](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=58800) early in the 4.8.x releases). Neither 4.8.4 or 4.8.5 (or RedHat's patched 4.8.2) trigger the issue. I'd consider this resolved.

mjuric:
@stuarteberg Also, I have a few minor patches I'll PR back to you -- I enabled multi-core builds in `build.sh` (`make -j$CPU_COUNT`) and had it add `cc -> gcc` symlink in `$PREFIX/bin` (there's code out there that explicitly looks for `cc` and will wind up using the system compiler otherwise).

jakirkham:
> Bottom line: the problem does not appear with 4.8.4 any more.

@mjuric, I am not sure that this is true. The problem was not with CentOS 5.x. It was an issue installing the prebuilt binary of this compiler on **other** distros and versions (e.g. Ubuntu, Debian, CentOS 6.x).

mjuric:
@jakirkham I was referring to https://github.com/conda/conda-recipes/pull/279#issuecomment-121382176

jakirkham:
Ok, sorry, I misunderstood we were on a new subthread. Carry on.

stuarteberg:
>I have a few minor patches I'll PR back to you

Thanks; just merged.

Glad to hear the issue seems to be fixed in gcc 4.8.4.

Now that I'm thinking about it, though, it still seems like it might be a good idea to add `--enable-checking=release` to `build.sh`, right?  (That's how gcc releases are typically configured.)  Since it's unrelated to portability issues, I've opened a separate PR for it:  #343.

frol:
Is there still work in progress? It sounds like the package is polished already. I would actually try to build gcc 5.1 mostly because I want to have a clear dependency for one of my projects that require libstdc++ and libgomp from 5.1. Is there already a work in this direction?

frol:
I've built gcc-5.2 on CentOS 5.11 (glibc 2.5, gcc-4.1). It works for me on that CentOS, Ubuntu 14.04, and Debian stable.

```
conda install -c salford_systems gcc
```

https://github.com/salford-systems/conda-recipes/tree/master/gcc-5.2

P.S. I used Docker CentOS 5.11 image with just a few packages from standart repo:

```
# Required to unpack GCC sources
$ yum install bzip2 tar

# Required by GCC
$ yum install gcc gcc-c++ zip

# Installed just in case, so please report if they are unnecessary
$ yum install perl unzip gzip
```

(I'm not sure if perl is required, and I also installed unzip, and gzip just in case after build failed because of lack of zip)

jakirkham:
@asmeurer, have you had a chance to test this yet?

frol:
There is an issue with my package. If you install it on an env where gcc run dependencies are not installed yet, it will fail on post-link.sh script as it will try to build "hello, world" tests. I don't feel like it is needed to be done there, so I'm removing that and rebuilding it...

stuarteberg:
>If you install it on an env where gcc run dependencies are not installed yet, it will fail on post-link.sh script

What run dependencies are you referring to?  Are referring to the dependencies listed in `meta.yaml` (`gmp`, `mpfr`, `mpc`, `isl`, `cloog`)?  Those should be pre-installed by conda before `post-link.sh` is executed.

jakirkham:
> If you install it on an env where gcc run dependencies are not installed yet, it will fail on post-link.sh script...

@frol, see ( https://github.com/conda/conda-recipes/pull/279#issuecomment-121266412 ).

stuarteberg:
@frol, what was the specific error you encountered?

jakirkham:
@stuarteberg @asmeurer, I took the liberty of rebasing this on the current master and put it here ( https://github.com/jakirkham/conda-recipes/tree/portable-gcc-4.8 ).

stuarteberg:
>I took the liberty of rebasing this on the current master

Good idea.  I just rebased and did a `git push -f` to this PR.  Now it can be auto-merged again.  (It was previously a non-auto-merge because of the build number.)

frol:
@stuarteberg @jakirkham My fail. It was indeed the system without system gcc installed... As to the issue with crt*.o, I checked Debian stable, Ubuntu 12.04/14.04, CentOS 5.11/6, Arch Linux and there are only two options really: /usr/lib/x86_64-linux-gnu/crt1.o (Debian-based distros) and /usr/lib64/crt1.o (others). Does it really worth the trouble of installing system-wide gcc for locating crt1.o?

stuarteberg:
>Is really worth the trouble of installing system-wide gcc for locating crt1.o?

This question has been [raised previously in this thread](https://github.com/conda/conda-recipes/pull/279#issuecomment-121266412).  My reasoning is that the `gcc` package is most likely to be used by software developers, who already have a system-wide gcc installed on their machine.  (Unlike the `libgcc` package, which is what a typical non-dev user will need.)

>I checked Debian stable, Ubuntu 12.04/14.04, CentOS 5.11/6, Arch Linux and there are only two options really

If those are really the only places where `crt*.o` might be found, then I guess we don't need the system gcc.  How sure of this are you?  Are there any other distros we should check?  (And is there an easy way to check them without needing to actually install them ourselves?)

frol:
@stuarteberg crt*.o is usually located in packages libc6-dev (Debian), glibc-devel (RedHat), glibc (Arch), so we need to inspect only these kind of packages.

jakirkham:
> Good idea. I just rebased and did a git push -f to this PR. Now it can be auto-merged again. (It was previously a non-auto-merge because of the build number.)

Excellent, deleted my copy of the branch.

jakirkham:
> ... crt*.o is usually located in packages libc6-dev (Debian), glibc-devel (RedHat), glibc (Arch), so we need to inspect only these kind of packages.

Though, I understand where you are coming from, not wanting to install the system compiler, @frol, I think @stuarteberg did the right thing here. He found a way to consistently find these object files no matter what system this `gcc` package is installed on. Every Linux platform makes it dead simple to install `gcc` if they don't just provide it out of the box. Unless you have an equivalent method to find these object files that does not result in a very long case statement that is likely to miss something, I am in favor of the existing solution.

frol:
@jakirkham How about using short case statement for popular locations (2 at this point) with backing with system gcc solution?

jakirkham:
@frol, to make sure I understand, you are suggesting check the 64-bit type directories, check the regular directories, and try `gcc` if all else fails. That sounds reasonable. In the best case, a system `gcc` is never needed. In the worst case, we get the exact same behavior. I feel that this searching should take into account the value of [`ARCH`]( http://conda.pydata.org/docs/building/environment-vars.html#environment-variables-set-during-the-build-process ).

frol:
@jakirkham I found even better solution. We can check in library search paths, which we can get with this trick:
```
$ /sbin/ldconfig -v 2>/dev/null | grep '^/'
```

Debian/Ubuntu output:
```
/usr/local/lib:
/lib/x86_64-linux-gnu:
/usr/lib/x86_64-linux-gnu: # <-- crt1.o
/lib:
/usr/lib: 
```

CentOS 5.11/6 output:
```
/lib:
/lib64:
/usr/lib:
/usr/lib64: # <-- crt1.o
/usr/lib64/sse2: (hwcap: 0x0000000004000000)
/usr/lib64/tls: (hwcap: 0x8000000000000000)
```

Arch output
```
/usr/lib/libfakeroot:
/usr/lib/julia:
/usr/lib32:
/usr/lib/openmpi:
/usr/lib: # <-- crt1.o (NOTE: /usr/lib64 is a symlink to /usr/lib and crt1.o is there)
```

frol:
@jakirkham And yes, I suggested to fallback to gcc in the worst scenario.

mjuric:
@frol Regarding `gcc-5.2` -- I think having it as an option would be useful, but I'd propose it to be a separate recipe (e.g., gcc-5.2, like there are [gcc-4.7](https://github.com/conda/conda-recipes/tree/master/gcc-4.7) and [gcc-4.9](https://github.com/conda/conda-recipes/tree/master/gcc-4.9)). It would also require a new `libgcc` (`libgcc-5.2`?). Perhaps open a separate PR?

I'd also worry about making `5.2` the default `gcc` just yet. It's still very bleeding edge (esp. with all the changes that went into `libstdc++` in the 5.x period); IMHO, it may be risky building the (conda) world with it just yet. Compiler bugs are notoriously evil to debug (or even to recognize)!

PS: It may be good to let this PR converge so @asmeurer can merge it -- it already brings considerable improvements!

frol:
@mjuric I agree that for some people 5.2 might be no mature enough, but we build our projects with 5.x for already almost 3 months and it works just fine for us. And I have already done libgcc-5.2 package also. I'll make a pull-request once this one is merged.

stuarteberg:
@frol, I implemented your idea to use `ldconfig` instead of the system `gcc` binary.  (See my most recent commit.)  I tested on Fedora-20 and Ubuntu-14.  I uploaded the new package to my binstar account:

    conda install -c stuarteberg gcc=4.8.4.99

So now this package doesn't require the user to have `gcc` installed on her machine.

jakirkham:
@stuarteberg, could we still use `gcc` as a fallback in case that doesn't work?

frol:
@stuarteberg @jakirkham That is cool. I have actually implemented it with fallback to GCC... Testing it right now...

frol:
@stuarteberg Let me finish testing my version and will see how to proceed.

mjuric:
@stuarteberg @frol Great idea about querying the linker! Would it [be better to query `ld`](http://stackoverflow.com/a/21610523) (the linker) rather than `ldconfig` (the *dynamic* linker). I.e., I think that this:
```
ld --verbose | grep SEARCH_DIR | tr -s ' ;' \\012 | cut -d \" -f 2
```
should give you the list of directories to search (tested only on RHELs, unfortunately). Example:
```bash
[mjuric@lsst-dev ~]$ ld --verbose | grep SEARCH_DIR | tr -s ' ;' \\012 | cut -d \" -f 2
/usr/x86_64-redhat-linux/lib64
/usr/local/lib64
/lib64
/usr/lib64
/usr/x86_64-redhat-linux/lib
/usr/lib64
/usr/local/lib
/lib
/usr/lib
```

frol:
@mjuric Debian output for `ld --verbose` is a little weird:

```

SEARCH_DIR("=/usr/x86_64-linux-gnu/lib64"); SEARCH_DIR("=/usr/local/lib/x86_64-linux-gnu"); SEARCH_DIR("=/usr/local/lib64"); SEARCH_DIR("=/lib/x86_64-linux-gnu"); SEARCH_DIR("=/lib64"); SEARCH_DIR("=/usr/lib/x86_64-linux-gnu"); SEARCH_DIR("=/usr/lib64"); SEARCH_DIR("=/usr/x86_64-linux-gnu/lib"); SEARCH_DIR("=/usr/local/lib"); SEARCH_DIR("=/lib"); SEARCH_DIR("=/usr/lib");
```

But indeed, this looks better! Thank you!

mjuric:
Sigh... :). Good catch. I guess one should be happy it's not _completely_ different :)...

jakirkham:
@froh, slight modification to make it a little more streamlined.

    /sbin/ldconfig -v 2>/dev/null | grep '^/' | cut -d : -f 1 | xargs -I {} find {} -name "crt[0123in].o" 2>/dev/null

stuarteberg:
>slight modification to make it a little more streamlined.

I agree with @mjuric -- it's better to use `ld` instead of `ldconfig`

frol:
@jakirkham I have actually came up with the following line:
```
$ ld --verbose | grep SEARCH_DIR | sed -r 's/SEARCH_DIR\("=?([^"]*)"\);/ \1/g'
```

I like oneliners, but not in scripts... :)

jakirkham:
@frol, what have you tested this line on thus far?

frol:
@jakirkham Works for me on CentOS, Debian, and Arch.

stuarteberg:
@frol, I'm ready to look at a pull request into this branch (stuarteberg:portable-gcc-4.8), whenever you're ready...

frol:
@stuarteberg Sure! I hope this will be the last iteration... Packing and unpacking gcc takes quite some time.

gcc-4.8/post-link.sh
frol: @asmeurer  This file is not only for OS X though. However, since you said that post-list.sh shouldn't be included on OS X. Could you please say why?

asmeurer: Is any of this process needed on OS X?

stuarteberg:
>Packing and unpacking gcc takes quite some time.

When I'm working on changes to `post-link.sh`, I usually test first using something like the following.  It helps avoid zipping/unzipping repeatedly....

```
cd conda-recipes/gcc-4.8
CONDA_ROOT=`conda info --root`
conda create -n test-gcc -c stuarteberg gcc=4.8.4.99

PREFIX=${CONDA_ROOT}/envs/test-gcc bash pre-unlink.sh
# Test my new version of post-link.sh
PREFIX=${CONDA_ROOT}/envs/test-gcc bash post-link.sh
```

gcc-4.8/post-link.sh
frol: @asmeurer The part below `fi` statement is kind of needed.

frol:
@stuarteberg I push a package to a fresh CentOS without gcc to test it properly... Nevertheless, thank you for the trick.

gcc-4.8/post-link.sh
asmeurer: Why is it needed? As far as I can tell, all it does is test the compiler. I haven't heard of any issues with the existing gcc package on OS X.

stuarteberg: This entire PR is really for Linux.  The things that this PR fixes ("fixed" std headers, inconsistent std header locations, and inconsistent locations for `crt0.o`, etc.) aren't issues on Mac.  If OSX had multiple "distributions" like Linux, that would be a different story.  It looks like we can just skip all this stuff on Mac.

>The part below fi statement is kind of needed.

The part below the last `fi` merely verifies that "Hello World" compiles.  Generally, it should always pass -- unless the user didn't install the xcode command-line tools on her machine.  (Otherwise gcc can't find the standard library headers.)

gcc-4.8/post-link.sh
```diff
@@ -0,0 +1,136 @@
...
+        
+        for INCDIR in ${EXTRA_SYSTEM_INCLUDE_DIRS}; do
+            # The following sed command will replace these two lines:
+            # *cpp:
+            # ... yada yada ...
+            #
+            # With these two lines:
+            # *cpp:
+            # ... yada yada ... -I${INCDIR}
+            sed -i ':a;N;$!ba;s|\(*cpp:\n[^\n]*\)|\1 -I'${INCDIR}'|g' ${SPECS_FILE}
```
asmeurer: I know there are compatibility problems with sed on OS X (see https://stackoverflow.com/questions/2320564/variations-of-sed-between-osx-and-gnu-linux). I am not sure if these ever come up with different Linux distributions as well. Might be worth investigating, since we want to keep this script as portable as possible.

gcc-4.8/post-link.sh
```diff
@@ -0,0 +1,136 @@
...
+    
+        # Remove the headers that gcc "fixed" as part of the gcc build process.
+        # They kill the gcc binary's portability to other systems,
+        #   and shouldn't be necessary on ANSI-compliant systems anyway.
+        # More discussion can be found here:
+        # https://groups.google.com/a/continuum.io/d/msg/conda/HwUazgD-hJ0/aofO0vD-MhcJ
+        while read x ; do
+          grep -q 'It has been auto-edited by fixincludes from' "${x}" \
+                   && rm -f "${x}"
+        done < <(find ${PREFIX}/lib/gcc/*/*/include*/ -name '*.h')
```
asmeurer: There is a syntax error here.

asmeurer:
You should run these through shellcheck to check for portability issues.

gcc-4.8/post-link.sh
jakirkham: Are you referring to the in-place portion? I haven't seen this with Linux, but this could always be changed to `-i.bak` or some other extension and that should work in all cases. Now, it would probably be a good idea to go back and clean these original files.

gcc-4.8/post-link.sh
stuarteberg: Weird.  I just copied that line from the gentoo gcc package:
https://sources.gentoo.org/cgi-bin/viewvc.cgi/gentoo-x86/eclass/toolchain.eclass?view=markup&sortby=log#l1614

... it seems to be working just fine for us.

jakirkham:
> You should run these through shellcheck to check for portability issues.

+1 Shellcheck is awesome!

gcc-4.8/post-link.sh
frol: I can confirm that this thing works while my vim checker also complains about this line. Even more, I cannot make it not to complain whatever I try. Pure magic... I will consider rewriting this.

gcc-4.8/post-link.sh
stuarteberg: I've run into the `-i` problem, too. But this script is intended to run on Linux only.  It has been tested on a variety of Linux distros at this point (see [OP](https://github.com/conda/conda-recipes/pull/279#issue-71995325) for details), so this `sed` syntax seems to be portable enough.

gcc-4.8/post-link.sh
asmeurer: Yeah, I'm going off of shellcheck. If it works, it's probably a bug in shellcheck.

stuarteberg:
>You should run these through shellcheck to check for portability issues.

Good idea.  Now that I'm looking, a lot of the complaints from shellcheck are due to the fact that I started the script with `#!/bin/sh` instead of `#!/bin/bash`.  I can assume conda will use `bash`, right?

asmeurer:
Yes, conda uses bash.

gcc-4.8/post-link.sh
asmeurer: The latest version of shellcheck gives this error message:

```
In gcc-4.8/post-link.sh line 28:
        done < <(find ${PREFIX}/lib/gcc/*/*/include*/ -name '*.h')
               ^-- SC2040: In sh, process substitution is not supported, even when sh is actually bash.
                      ^-- SC2086: Double quote to prevent globbing and word splitting.
```

stuarteberg:
Now shellcheck is happy.

gcc-4.8/post-link.sh
stuarteberg: That error disappeared after I changed the first line from `#!/bin/sh` to `#!/bin/bash`.  I pushed a new commit just now, and shellcheck is happy.

frol:
@stuarteberg I have pull-requested my changes. Please, review them. Unfortunately, now as you fixed some quotings also it won't merge clearly. Please, grab just the part for "issue_#2" then.

stuarteberg:
>I have pull-requested my changes. Please, review them.

Merged!

frol:
@stuarteberg Great!
BTW, I've just pushed new gcc-5 package! In a little while I will also update libgcc-5 and pull-request both.

stuarteberg:
BTW, I just updated my binstar channel with the latest version (4.8.4.99, build number 5).

frol:
@stuarteberg I'm testing my gcc-5 package (built on CentOS 5.11) on CentOS 6.6 and it fails in some cases (not on "hello, world" examples though) with the following error:

```
In file included from /usr/local/miniconda/pkgs/gcc-5-5.2.0-1/include/c++/5.2.0/bits/localefwd.h:40:0,
                 from /usr/local/miniconda/pkgs/gcc-5-5.2.0-1/include/c++/5.2.0/string:43,
                 from /usr/local/miniconda/pkgs/gcc-5-5.2.0-1/include/c++/5.2.0/stdexcept:39,
                 from ../license_management/libsslicense/libsslicense/include/sslicense/LicenseProvider.h:12,
                 from ../license_management/libsslicense/libsslicense/include/sslicense/sslicense.h:11,
                 from ../license_management/libsslicense/libsslicense/src/stdafx/stdafx.h:7,
                 from ../license_management/libsslicense/libsslicense/src/stdafx/stdafx.cpp:2:
/usr/local/miniconda/pkgs/gcc-5-5.2.0-1/include/c++/5.2.0/x86_64-unknown-linux-gnu/bits/c++locale.h:52:23: error: ‘uselocale’ was not declared in this scope   extern "C" __typeof(uselocale) __uselocale;
```

It is already the second issue though it looks exactly like the one I have already patched with a solution that I've found in mail lists (https://gcc.gnu.org/ml/gcc-help/2011-01/msg00325.html) - I defined `__USE_XOPEN2K8` in `./lib/gcc/x86_64-unknown-linux-gnu/5.2.0/include-fixed/features.h`

Have you seen such behaviour?

NOTE: I have successfully built `msgpack` and `gtest` with the hack of defining `__USE_XOPEN2K8`, but I cannot build our proprietary project

stuarteberg:
@frol, That error occurs when you attempt to use a copy of `features.h` that isn't intended for your OS.  In this case, `features.h` came from CentOS 5.11, where the gcc package was created.

To avoid this error, `post-link.sh` is supposed to delete the "fixed" version of `features.h` from your environment at *install* time.  That means you must always use `gcc` from an actual environment.  From the error you posted above, it seems like you're attempting to use gcc (or its headers, at least) directly from the `pkgs` directory in your miniconda root directory, without installing it first.

frol:
@stuarteberg I've installed it to the root env. `features.h` doesn't get deleted because it does not contain "auto-edited"! However, I figured that I need to add /usr/include implicitly to get the proper `features.h`. In fact, if there is no `features.h` at all (I moved all occurences out from the system) gcc doesn't complain about missing headers! Weird.

stuarteberg:
>I've installed it to the root env.

Still: In that case I would expect your error message to begin with:

    In file included from /usr/local/miniconda/include/c++/5.2.0/bits/localefwd.h:40:0,

not this:

    In file included from /usr/local/miniconda/pkgs/gcc-5-5.2.0-1/include/c++/5.2.0/bits/localefwd.h:40:0,
    										   ^^^^^^^^^^^^^^^^^^^

>`features.h` doesn't get deleted because it does not contain "auto-edited"!

Ah-ha.  So you need to fix the `grep` line in step 1 of `post-link.sh`.  Our method for identifying "fixed" (i.e. patched) std library headers is apparently not valid any more for gcc 5.

>In fact, if there is no features.h at all (I moved all occurences out from the system) gcc doesn't complain about missing headers! Weird.

You removed the *system* copy `features.h` (`/usr/include/features.h`)?  That is weird.  gcc must be locating a version of it somewhere else... I suppose you could figure it out via `gcc -v test.c`  Anyway, it won't matter once you fix `post-link.sh` so that the bad `features.h` gets deleted properly.

frol:
@stuarteberg I guess the paths are weird is because I patched conda build to use symlinks always because of this bug: https://github.com/conda/conda/pull/1410 (the pull-request fixes it properly, but I use `sed` to make a one-line patch)

I figured that I removed system `features.h` and the one from the root env, but I forgot about that I'm using local env and it had a link to `pkgs`...

frol:
@stuarteberg Indeed, once I removed features.h from the package everything got fixed! Looking into `post-link.sh` fix now...

frol:
@stuarteberg My fail again... It was because of the package renaming to `gcc-5`, I didn't rename the link scripts accordingly. Uploading 5.2.0-2...

frol:
Uploaded `gcc-5` and `libgcc-5` (version 5.2.0):

```
$ conda install -c salford_systems gcc-5
```

```
$ conda install -c salford_systems libgcc-5
```

It works for me now!

frol:
@stuarteberg BTW, because of the issue in conda (https://github.com/conda/conda/issues/1451) our post-link.sh fails on "Hello, World" when using `conda create --clone ...`.

asmeurer:
So perhaps some files here need to be set as [no link](http://conda.pydata.org/docs/building/meta-yaml.html#no-link).

stuarteberg:
>So perhaps some files here need to be set as no link.

Sorry, I'm missing something.  Which files?  Why?

FWIW, I don't think the issue @frol mentioned above (conda/conda#1451) would be helped in any way by `no_link`, does it?  Were you referring to some other issue?

asmeurer:
It sounded like there were issues with following symlinks into the pkgs directory.

stuarteberg:
>It sounded like there were issues with following symlinks into the pkgs directory.

I don't think there are any symlink issues.  (Earlier in this thread, I had a moment of confusion due to @frol's use of symlinks in a special branch of conda he's been hacking on, but it was a red herring.  The symlinks were not to blame.)

stuarteberg:
As far as I know, this PR has no outstanding problems.  It's ready to be merged.

mjuric:
@stuarteberg @asmeurer I can confirm the code from the PR worked beautifully (allowed us to build gcc on CentOS5 and use it on CentOS5 and 6 machines).

gcc-4.8/post-link.sh
```diff
@@ -0,0 +1,170 @@
...
+            # *cpp:
+            # ... yada yada ...
+            #
+            # With these two lines:
+            # *cpp:
+            # ... yada yada ... -I${INCDIR}
+            sed -i ':a;N;$!ba;s|\(*cpp:\n[^\n]*\)|\1 -I'${INCDIR}'|g' "${SPECS_FILE}"
+        done
+    fi
+fi
```
asmeurer: Let's put the whole file in this block.

izaid:
I'm giving this PR a massive +1. DyND would not build until I found this PR.

gcc-4.8/post-link.sh
stuarteberg: >Let's put the whole file in this block.

OK, on Darwin, the post-link script now exits immediately. (@asmeurer, Sorry for the delay; I was on vacation.)

gcc-4.8/pre-unlink.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
asmeurer: Shouldn't this script also exit immediately on OS X?

asmeurer:
Should we go ahead and update to 4.8.5?

gcc-4.8/pre-unlink.sh
stuarteberg: OK, done.

stuarteberg:
>Should we go ahead and update to 4.8.5?

My original intent was to limit the scope of this PR to fixing portability issues.  Personally, I think it's cleaner to upgrade to 4.8.5 via a separate PR.  But if it's easier for you if I change it here, I can do so.

asmeurer:
I'd rather only build this once if possible, so yeah, go ahead and updated it. Then I'll merge this and see if I can build it (and let you know if I have issues).

stuarteberg:
>yeah, go ahead and update it.

OK, done (via 52dc383).  AFAICT, no dependency versions need to change (for cloog, gmp, mpfr, etc.)

asmeurer:
Great. It'll obviously take a while, but I'll try building this and report back.

jakirkham:
:tada:

orodbhen:
I tested this by trying to build FFMpeg with a bunch of optional dependencies on OS X .Yosemite. Most of the packages built fine, but it failed on libvpx with a bunch of errors about invalid processor instructions. Here's a snippet:

    /var/tmp//cct7Y7nM.s:967:no such instruction: `vmovq %xmm2, (%r15)'
    /var/tmp//cct7Y7nM.s:968:no such instruction: `vmovhps %xmm2, (%rdx)'
    /var/tmp//cct7Y7nM.s:969:no such instruction: `vpand 120(%rsp), %xmm1,%xmm1'
    /var/tmp//cct7Y7nM.s:970:no such instruction: `vpor %xmm1, %xmm9,%xmm1'  
    /var/tmp//cct7Y7nM.s:971:no such instruction: `vmovq %xmm1, (%rax)'
    /var/tmp//cct7Y7nM.s:972:no such instruction: `vmovhps %xmm1, (%rdi)'
    make[1]: *** [vp9/common/x86/vp9_loopfilter_intrin_avx2.c.o] Error 1
    make: *** [.DEFAULT] Error 2

Conda built the package successfully, however, when using the compiler bundled with XCode.

frol:
@orodbhen My answer is not exactly relevant, but I've seen something similar while building Boost with my `gcc-5` package on Linux. In my case, the issue was an old linker (`ld` from `binutils` package), so I installed `binutils` from Anaconda.org.

jakirkham:
@orodbhen, this recipe doesn't really change how the GNU compilers are built or installed on Mac. It only changes how GNU compilers are installed on Linux. Maybe it would be better for you to open a new issue so it can attract some needed attention.

gcc-4.8/build.sh
```diff
@@ -1,33 +1,41 @@
...
-rm $PREFIX/lib64
+rm "$PREFIX"/lib64
+
+# Link cc to gcc
+(cd "$PREFIX"/bin && ln -s gcc cc)
+
+# For reference during post-link.sh, record some
+# details about the OS this binary was produced with.
+mkdir -p "${PREFIX}/share"
+cat /etc/*-release > "${PREFIX}/share/conda-gcc-build-machine-os-details"
```
asmeurer: Well this is frustrating. The OS X build got all the way here and then failed.

stuarteberg: Aw, crap. Sorry.  I guess these lines should be inside the Linux path.  I see you've already changed it.

asmeurer:
I built this on CentOS 5 for 64-bit and 32-bit Linux, and it failed to install on 32-bit Ubuntu 14.04.2 LTS

```
In file included from /usr/include/stdio.h:27:0,######################################################            |  87%
                 from hello.c:1:
/usr/include/features.h:374:25: fatal error: sys/cdefs.h: No such file or directory
 #  include <sys/cdefs.h>
                         ^
compilation terminated.
Installation failed: gcc is not able to compile a simple 'Hello, World' program.
Error: Error: post-link failed for: gcc-4.8.5-0
```

The 64-bit version seems to work. Both builds are on my `asmeurer` binstar channel.

stuarteberg:
>it failed to install on 32-bit Ubuntu

Ah, sorry, I didn't test on any 32-bit systems.

The problem is almost certainly that we're missing an include path here:
https://github.com/conda/conda-recipes/blob/master/gcc-4.8/post-link.sh#L102

If you change that line from:

    EXTRA_SYSTEM_INCLUDE_DIRS="/usr/include/x86_64-linux-gnu"

to:

    EXTRA_SYSTEM_INCLUDE_DIRS="/usr/include/x86_64-linux-gnu /usr/include/i686-linux-gnu"

or maybe:

    EXTRA_SYSTEM_INCLUDE_DIRS="/usr/include/$(uname -m)-linux-gnu"

... then I think it will work.  Obviously, you'll save time if you skip rebuilding the package and simply edit `pkgs/linux-32_gcc-4.8.5-0/bin/.gcc-post-link.sh` directly.  Then try installing gcc again and see if it works.

asmeurer:
`uname -m` is `i686` but the directory is `i386-linux-gnu`. I don't know if that's always the case. Maybe I should check for both.

stuarteberg:
>`uname -m` is `i686` but the directory is `i386-linux-gnu`

How annoying. :-/

>Maybe I should check for both.

Sounds good.  Also, we should check for non-existent directories. If you haven't already done so, I think the following code should do the trick.

```bash
    EXTRA_SYSTEM_INCLUDE_DIRS="/usr/include/x86_64-linux-gnu /usr/include/i386-linux-gnu /usr/include/i686-linux-gnu"
    
    for INCDIR in ${EXTRA_SYSTEM_INCLUDE_DIRS}; do
        if [ ! -d "${INCDIR}" ]; then
            continue
        fi
        
        # The following sed command will replace these two lines:
        # *cpp:
        # ... yada yada ...
        #
        # With these two lines:
        # *cpp:
        # ... yada yada ... -I${INCDIR}
        sed -i ':a;N;$!ba;s|\(*cpp:\n[^\n]*\)|\1 -I'${INCDIR}'|g' "${SPECS_FILE}"
    done
```

jakirkham:
@stuarteberg, should `i686` proceed `i386` in the list or does it matter?

stuarteberg:
>should i686 proceed i386 in the list or does it matter?

Sigh.  I'm not sure.  I guess the best thing would be to check the system gcc on the 32-bit VM.  If both directories are listed, then we should use the same order it uses.  This command will show the header search directories (among other things):

    $ echo | gcc -v -x c -E -

I don't happen to have a 32-bit Ubuntu VM handy, so I can't test this myself.

asmeurer:
[This](https://github.com/conda/conda-recipes/commit/79e032b5e8cff84486409157e17e0f04516c23ba) is what I did.

stuarteberg:
>[This](https://github.com/conda/conda-recipes/commit/79e032b5e8cff84486409157e17e0f04516c23ba) is what I did.

That should be just fine.  Have you tested it yet?

FWIW, I just spun up a new Ubuntu 32-bit VM just to check the include path order  Only `/usr/include/i386-linux-gnu` appears in the search path.  I think there's no harm in having the other directory listed, though.

asmeurer:
It seems to install and remove fine. I haven't tested it beyond that, though. The builds are on my binstar.

izaid:
@asmeurer Can we merge it then? Or are we waiting for something else?

stuarteberg:
>Can we merge it then? Or are we waiting for something else?

He already merged it into conda-recipes master.  The last step would be to add it to the default conda package repo, so it can be installed via `conda install gcc` instead of `conda install -c asmeurer gcc`

izaid:
Yes, sorry, that's what I meant. @asmeurer is this something we can do?

asmeurer:
I've also put it in on the `r` channel.

izaid:
@asmeurer Is this something that we can get added to the default conda repo? Or is that unlikely to happen for whatever reason?

asmeurer:
It will happen. I just want it to get tested a little bit before doing so.

izaid:
Great, thanks!

Question: Is it worth making a gcc-libraries recipe that simply grabs the important libraries (libstdc++, ...) from gcc? This way other recipes can depend on the gcc libraries without needing all of GCC.

I've done this for DyND and it works nicely.

jakirkham:
@izaid, I believe this is already done. ( https://github.com/conda/conda-recipes/blob/master/libgcc/meta.yaml )

izaid:
@jakirkham Ah, okay, good.

But, don't we need another one for the gcc-4.8 related to this PR?

jakirkham:
I have PR for bumping the version to be consistent with the version of `gcc` here. ( https://github.com/conda/conda-recipes/pull/371 )

jakirkham:
Also, @izaid, there is a copy of `libgcc` 4.8.4 in the anaconda channel that @asmeurer had added. ( https://anaconda.org/anaconda/libgcc ).

asmeurer:
oh I forgot about libgcc. I should rebuild that as well.

dfroger:
Hello,

It may be usefull to create a symbolic link `cc` to `gcc`, otherwise `CMake` (don't know for other build system) choose `/usr/bin/cc` instead of `gcc` of the conda package.

For C++, it seems ok (`CMake` choose `c++` of the conda package).

asmeurer:
That sounds like a good idea, although I'm wondering if that would break some of my OS X recipes that are currently working with clang.

stuarteberg:
>It may be usefull to create a symbolic link cc to gcc

Already done: @mjuric already did that in 510897e2dac434e774b33a9afdc391c8dd556db8, which was merged into this PR before it was accepted by @asmeurer.

@dfroger, I'm guessing you are using `conda install -c stuarteberg gcc=4.8.4.99`, which I created before the @mjuric provided the `cc` fix.  Sorry about that.  Please try `conda install -c asmeurer gcc=4.8.5` instead.  It contains everything from this PR, plus more.  (Hopefully his version will become the official `gcc` conda package soon...)

>I'm wondering if that would break some of my OS X recipes that are currently working with clang

If you don't include the `gcc` package as a build dependency, then there should be no problems, right?  If `gcc` isn't installed, it won't be on your path (and neither will the new `cc` symlink), and CMake won't find it.

asmeurer:
I guess I am primarily thinking of R but that is managed via https://github.com/conda/conda-recipes/blob/7e78b47686c193fb81973d33b3797ed259eaccdf/r-base/build.sh#L61.

dfroger:
@stuarteberg exactly, I was using`conda install -c stuarteberg gcc=4.8.4.99`, I'll switch to `conda install -c asmeurer gcc=4.8.5`. Thanks a lot!

stuarteberg:
Yes, it's best for all recipes to be explicit about which compiler they're using, especially if they involve C++ code.   That means explicitly setting the `CC` and `CXX` environment variables, and also explicitly setting `CMAKE_C_COMPILER` and `CMAKE_CXX_COMPILER` when using cmake.

For my own project, the above guidelines are spelled out here: https://github.com/ilastik/ilastik-build-conda#appendix-compiler-details

dfroger:
Thanks for the guidelines. By the way, I'm using this Vagrant box: https://github.com/dfroger/conda-build-env

PR #282, opened 2015-05-02
Update mercurial to 3.4

ngoldbaum:
2.7 is almost two years old at this point, so let's upgrade to the latest and greatest.

http://mercurial.selenic.com/wiki/WhatsNew
http://mercurial.selenic.com/wiki/UpgradingMercurial


ccordoba12:

PR #283, opened 2015-05-04
Add recipe for coursera-dl

hargup:
[Coursera-dl](https://github.com/coursera-dl/coursera) is a popular tool to download course content from coursera.org courses.

### DIFF
--- a/coursera-dl/meta.yaml
+++ b/coursera-dl/meta.yaml
@@ -1,0 +1,68 @@
+package:
+  name: coursera-dl
+  version: "3.1.1"
+
+source:
+  fn: coursera-dl-3.1.1.tar.gz
+  url: https://pypi.python.org/packages/source/c/coursera-dl/coursera-dl-3.1.1.tar.gz
+  md5: c2a5089ee5b78874f1f4a9519fb115a8
+#  patches:
+   # List any patch files here
+   # - fix.patch
+
+build:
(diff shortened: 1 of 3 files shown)


coursera-dl/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+  # If this is a new build for the same version, increment the build
+  # number. If you do not include this key, it defaults to 0.
+  # number: 1
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - beautifulsoup4
```
ccordoba12: This package is called `beautiful-soup` in conda

coursera-dl/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - mechanize
```
ccordoba12: `mechanize` doesn't exist in `conda-recipes`, so you'd need a recipe for it too.

coursera-dl/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - argparse
```
ccordoba12: `argparse` is part of the standard library in Python 2.7, 3.3 and 3.4. So please remove this dependency.

groutr: or make it a dependency only for Python 2.6 with `- argparse     [py26]`

PR #285, opened 2015-05-04
vtk: Fixed build.sh: Use "sed -i" instead of non-existent $REPLACE command

stuarteberg:
vtk: Fixed build.sh: Use "sed -i" instead of non-existent $REPLACE command


ccordoba12:
Thanks! Merging :-)

PR #287, opened 2015-05-04
vtk: Build with Qt and PyQt extensions.

stuarteberg:
I'm not 100% sure if this PR is ready to be accepted as-is.  I'm hoping for some guidance.

Background: VTK can optionally be configured to build special Qt-based extensions, and furthermore, one can build PyQt (sip) bindings for them.  The current version of the Anaconda `vtk` package does not include the Qt extensions.

Fortunately, adding them to the build is easy.  My only question is about how to express the dependencies properly in `meta.yaml`.  For now, I simply made `qt` and `pyqt` **build** dependencies, but omitted them from the **run** dependencies.  This way, users who don't need vtk-qt interop don't need to install qt.  (They'll have an extra file in their `lib` directory (`libvtkQtPythonD.so`), but so what?) But users who _do_ want the interop can simply install `vtk` and `pyqt` separately, and then everything works just fine.

If I understand correctly, this is the sort of situation a conda "feature" might be used for.  But I don't know exactly how that works.  Is it really necessary to do that in this case, or is this simple solution good enough?

**Edit:** See also: https://groups.google.com/a/continuum.io/d/msg/anaconda/Q0XyziKE-f8/bkoQN2FtBQAJ


vtk/meta.yaml
```diff
@@ -12,12 +12,14 @@ source:
...
 build:
-  number: 1                                  [linux]
+  number: 2                                  [linux]
 
 requirements:
   build:
     - cmake                                  [linux]
     - python
+    - qt
+    - pyqt
```
ccordoba12: Only `pyqt` is necessary here because it'll bring `qt` as a dependency.

vtk/run_test.py
```diff
@@ -12,3 +12,8 @@
 import vtk.vtkViewsPython
 import vtk.vtkVolumeRenderingPython
 import vtk.vtkWidgetsPython
+
+# Qt extension
+import vtk.QVTKWidget
+import vtk.QVTKWidget2
+import vtk.vtkQImageToImageSource
```
ccordoba12: if `pyqt` is not added as a runtime dependency, then I think these tests are going to fail.

stuarteberg: Whoops, nice catch.  I'll add `pyqt` as a test dependency.

ccordoba12:
I don't know what to answer about your question.

@asmeurer, do you have something to say?

vtk/meta.yaml
stuarteberg: Yeah, it's a matter of style.  Technically, I'm adding two extensions to the VTK build: One needs Qt (only), and the other needs PyQt (and Qt).  I suppose an analogy would be if I had a Python C-extension module that used the Python HDF5 API **and** also used h5py in a few places.  Would I list only `h5py` in that case?

ccordoba12: I'm just saying that, as far as conda is involved, adding `pyqt` is enough because that dependency will pull `qt` by default.

stuarteberg: OK, done.

frol:
It seems just like matplotlib case: the built package can work without Qt, but also provide extended features when Qt is available. I want to understand whether there is an issue with creating `-qt`-suffixed meta-package with `run: - pyqt` requirement.

ccordoba12:
@stuarteberg, I'm in charge of VTK now, and I think if we want this functionality, `pyqt` should be also a runtime dep.

Else, it'd very hard for people to discover that VTK has support for Qt. What do you think?

stuarteberg:
>I'm in charge of VTK now, and I think if we want this functionality, pyqt should be also a runtime dep.

Qt is a somewhat heavy dependency (about the same as VTK).  So adding that qt dependency for people who don't need it doubles the size-on-disk from ~170 MB to ~340 MB.  I guess that's not a huge deal, but it seems unnecessary.

>Else, it'd very hard for people to discover that VTK has support for Qt.

I think most people would discover the VTK-Qt by browsing VTK's documentation.  When they try it for the first time, they'll get a an import error complaining about Qt not being available.  I think the natural thing to do in that case would be to say "hmm, maybe I should install Qt and try again..."

But I don't know.  You have way more experience responding to questions from typical conda users, so I could be wrong.  I certainly won't complain no matter which decision you make.

stuarteberg:
On an unrelated note: this VTK recipe is for VTK 5.  VTK 6 introduces breaking changes (and drops support for the PyQt bindings altogether, I think...).  If you ever upgrade this recipe to VTK6, I think VTK 5 users will appreciate continued maintenance for this old version.

PR #288, opened 2015-05-06
ENH : make pkg-config work outside of a _build env

tacaswell:


ccordoba12:
Could you explain your rationale for this change?

tacaswell:
Without this change conda does not re-write the hard-coded paths inside the `pkg-config` binary which set where it looks for `*.pc` files (in my case `/home/tcaswell/.pe/_build/lib`).  With out this change running `pkg-config` in any environment not at exactly the same place as the env it was built it,   With this change `pkg-config` will work correctly for any environment it is installed into.  I am not really sure why this hasn't been an issue before, I am guessing some mix of this was mostly used on the machine that built it (this isn't in the main channel, just @asmeurer s so I assume this is part of his build system for the more persnickety packages) and things hitting system pkg-config and the base system happening to have the correct libraries.

ccordoba12:
Thanks for the explanation.

PR #290, opened 2015-05-11
DOC: README.md: add headings, update links

westurner:


README.md
```diff
@@ -1,17 +1,100 @@
 # Conda Recipes
 
-Example recipes for the new conda build system.  Use
+Community-maintained repository of conda package build recipes
+for 592+ conda packages.
```
ccordoba12: I don't agree with putting a specific number here but it'll get outdated quite quickly

README.md
```diff
@@ -1,17 +1,100 @@
...
-    conda build <recipe directory>
+* Package maintainers SHOULD upgrade packages
+  when appropriate.
 
-You have to use conda >= 1.7
-[conda](https://github.com/continuumio/conda).
+* Conda recipes MAY run other scripts:
 
-See http://docs.continuum.io/conda/build.html for information on how to make a recipe,
-or just look at the examples here.
```
ccordoba12: Why removing the link to our docs? I also don't agree with it :-)

ccordoba12: Sorry, I just saw that you added an updated link below

westurner: Still there in line 36: https://github.com/conda/conda-recipes/pull/290/files#diff-04c6e90faac2675aa89e2176d2eec7d8R36 :0)

README.md
westurner: How about "hundreds of packages"?

README.md
```diff
@@ -1,17 +1,100 @@
...
+Community-maintained repository of conda package build recipes
+for 592+ conda packages.
 
-    conda build <recipe directory>
+* Package maintainers SHOULD upgrade packages
+  when appropriate.
 
-You have to use conda >= 1.7
-[conda](https://github.com/continuumio/conda).
+* Conda recipes MAY run other scripts:
```
ccordoba12: I don't understand this MAY here. Doesn't conda always run build.sh or build.bat?

README.md
ccordoba12: Sure, fine by me :-)

README.md
```diff
@@ -1,17 +1,100 @@
...
-See http://docs.continuum.io/conda/build.html for information on how to make a recipe,
-or just look at the examples here.
 
-This project (`conda-recipes`) is in the public domain.   Note that this statement
+  * at build time by specifying ``script:`` in ``meta.yaml``
+    (default: ``build.bat`` / ``build.sh``)
+  * at build time in `build.bat` (Windows) and `build.sh` (OSX, Linux)
+  * before the package is installed in ``pre-link.sh``
+  * after the package is installed in ``post-link.sh``
+  * before the package is removed in ``pre-unlink.sh``
```
ccordoba12: I think the optional scripts are these ones (pre-link. post-link and pre-unlink)

README.md
westurner: If ``scripts:`` is set, that overides the default ``build.bat`` and ``build.sh`` scripts.

I could nest the "default:" part

README.md
westurner: I'll prefix with ``(optional)``

README.md
```diff
@@ -1,17 +1,100 @@
...
+    conda skeleton pypi rdflib-jsonld
+
+    # Generate a conda recipe from CRAN (R)
+    cd conda-recipes/r-packages
+    conda skeleton cran r-chron
+    conda skeleton cran r-dplyr
+
+    # Generate a conda recipe from CPAN (Perl)
+    conda skeleton cpan DateTime
+    conda skeleton cpan RDF::Helper
```
ccordoba12: I think you could explain a bit more what `skeleton` does in this section. I mean, you could say that users can automatically generate recipes for simple Python, R and Perl packages by using `skeleton`, and then give the examples below.

README.md
westurner: Aren't build.bat and build.sh sort-of optional as well?

ccordoba12: Ok

README.md
ccordoba12: I see, but that's too much detail for a Readme, which IMHO should explain the big picture :-)

README.md
ccordoba12: But only if you set `script:`. I'd say they are required, at least in the big picture of conda-build

README.md
```diff
@@ -1,17 +1,100 @@
...
+  * before the package is installed in ``pre-link.sh``
+  * after the package is installed in ``post-link.sh``
+  * before the package is removed in ``pre-unlink.sh``
+
+  http://conda.pydata.org/docs/build.html
+
+* Some of these recipes do intersect with
+  [Conda packages included with Anaconda](http://docs.continuum.io/anaconda/pkg-docs.html).
+
+* You can [search for Conda packages hosted in free public BinStar Channels](https://binstar.org/search?q=).
```
asmeurer: It's just Binstar (not BinStar).

README.md
asmeurer: I agree with @ccordoba12

README.md
```diff
@@ -1,17 +1,100 @@
...
+
+## Building Conda Recipes
+To build conda packages from a conda recipe:
+
+```bash
+git clone ssh://git@github.com/conda/conda-recipes
+recipe_name="conda"
+conda build "conda-recipes/${recipe_name}"
+ls -al ${CONDA_ROOT}/conda-bld/** | grep "${recipe_name}"
+```
```
ccordoba12: This seems overly complicated. What about:

```bash
git clone ssh://git@github.com/conda/conda-recipes
cd conda-recipes
conda build matplotlib
```

And then explaining that the created conda package is (typically) located at

    ~/{Anaconda,miniconda}/conda-bld/arch/

where `arch` can be linux-32, linux-64, etc

README.md
```diff
@@ -1,17 +1,93 @@
...
+though a project `foo` might have an MIT, Apache,
+[or any other license](http://choosealicense.com/licenses/),
+the recipe for project `foo` (within this repository) is Public Domain.
+
+* The [`conda skeleton`](http://conda.pydata.org/docs/commands/build/conda-skeleton.html)
+commands copy any declared package license metadata into the `meta.yaml` file.
+
+* Package creators and maintainers SHOULD include license information
+in their respective ``meta.yaml`` files. For example, from
+[conda-recipes/conda/meta.yaml](https://github.com/conda/conda-recipes/blob/master/conda/meta.yaml)
```
ccordoba12: Could you use a different package other than `conda` for this kind of examples? I think It's a bit confusing :-)

westurner: Trying to avoid picking any favorites. Maybe a recipe for ``conda-build``?

The conda recipe is very simple, shows git usage, and is very simple.

ccordoba12: You could a really know package, like `matplotlib`. In any case, you're showing other examples of git recipes above.

README.md
westurner: I added ``recipe_name`` as a path containing a '/' to demonstrate that conda-build accepts a path (there is no magic).

```bash
git clone ssh://git@github.com/conda/conda-recipes
conda build conda-recipes/conda
# ...
```

I added the CONDA_ROOT variable because different installations are different. (e.g. ``/opt/conda``, ``~/-wrk/-conda27`` (https://github.com/westurner/dotfiles/blob/release/0.8.3/etc/bash/08-bashrc.conda.sh).

I could list a number of CONDA_ROOT examples.

I think it's easier to understand that the prefix is CONDA_ROOT. I find this confusing:

    ~/{Anaconda,miniconda}/conda-bld/arch/

README.md
westurner: There's a link to the skeleton docs in #license, I'll duplicate that here.

README.md
westurner: A ``summary:`` attr would be good.

westurner:
#TODO

PR #296, opened 2015-05-11
Delete freeimage/bld.bat as it's using setup.py for a make build

jakirkham:
Building goes through `make` not `python`.


jakirkham:
Sorry for all the noise. Trying to find the right way to describe this without making it sound scary or weird, which it isn't.

ccordoba12:
Ok, no problem :-)

PR #298, opened 2015-05-12
[WIP] Split GDAL package into libgdal and gdal

groutr:
This will split gdal into libgdal (python independent) and gdal (python bindings).  All packages are built from source on each of the platforms.

TODO:
- [ ] Fix build errors on OS X (failing on my system with an error about libiconv)
- [x] Strip binaries on linux (they are huge, stripping them reduces their size ~75%)

Feedback/testing welcome.

### DIFF
--- a/gdal/bld.bat
+++ b/gdal/bld.bat
@@ -7,13 +7,9 @@ if errorlevel 1 exit 1
 %PYTHON% setup.py build_scripts
 if errorlevel 1 exit 1
 
-%PYTHON% setup.py install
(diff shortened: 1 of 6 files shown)


jakirkham:
+1

timcera:
Minor detail and perhaps an unconventional solution, but the names 'gdal-libs' and 'gdal-python' would sort next to each other.

jlaura:
Running into the libiconv issue on OS X right now trying to build with this recipe.  Any status / suggestions on getting this to build on OS X?

PR #313, opened 2015-05-28
tifffile: Build from scratch.

jakirkham:
Tested on Mac OS 10.9 and CentOS 6.3. Unable to test on Windows, but expect it should work.

### DIFF
--- a/tifffile/meta.yaml
+++ b/tifffile/meta.yaml
@@ -1,0 +1,65 @@
+package:
+  name: tifffile
+  version: "0.4"
+
+source:
+  fn: tifffile-0.4.tar.gz
+  url: https://pypi.python.org/packages/source/t/tifffile/tifffile-0.4.tar.gz
+  md5: d92dba2d7ea08e47bfbf9282c2d2e8dc
+#  patches:
+   # List any patch files here
+   # - fix.patch
+
+# build:
+  # preserve_egg_dir: True
+  # entry_points:
+    # Put any entry points (scripts to be generated automatically) here. The
+    # syntax is module:function.  For example
+    #
+    # - tifffile = tifffile:main
+    #
+    # Would create an entry point called tifffile that calls tifffile.main()
+
+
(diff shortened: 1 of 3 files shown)


tifffile/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    #
+    # Would create an entry point called tifffile that calls tifffile.main()
+
+
+  # If this is a new build for the same version, increment the build
+  # number. If you do not include this key, it defaults to 0.
+  # number: 1
+
+requirements:
+  build:
```
stuarteberg: These requirements aren't right:

- IIUC, `libtiff` is not a dependency of `tifffile` (not even an optional dependency).
- The `tifffile.py` docstring is specific about version requirements:
  - `python 2.7*|>=3.4`
  - `numpy >=1.8.2`
- `matplotlib` is an optional dependency.  I think it should not be listed here. User who need plotting can install it separately.

stuarteberg:
This PR is a dupe of #195, which also lacks proper version requirements.

jakirkham:
Actually, I am going to close this as this has been included in `skimage`. See code here ( https://github.com/scikit-image/scikit-image/tree/master/skimage/external/tifffile ) and API ( http://scikit-image.org/docs/dev/api/skimage.external.tifffile.html ). Note they simply include tifffile directly so it is not even a dependency.

jakirkham:
@stuarteberg I don't follow your complaint regarding #195. It appears that they have included `numpy` and `python` without anything additional. Is it simply that you are wishing to see version strings? Regardless, I think this is moot point.

jakirkham:
Sorry of the dupe.

stuarteberg:
>this has been included in skimage.

Excellent.  Good to know, thanks.

Regarding #195:

>Is it simply that you are wishing to see version strings?

Yes.

jakirkham:
To be honest, I think I ran `conda skeleton tifffile` and just committed it.

PR #316, opened 2015-06-02
Add missing GCC libraries to libgcc (Java, Objective-C)

jakirkham:


asmeurer:
These are really big files. If these are really needed, we should probably use a separate package.

ccordoba12:
I also think they should be added to a different package. They are not really necessary for C/C++ programs.

jakirkham:
Unfortunately, separating them is not really feasible as they are all linked to `libgcc_s.1.dylib` or on linux  `libgcc_s.so.1`.

asmeurer:
The other package could depend on libgcc.

jakirkham:
Ok, we could do it that way. As a thought, it would also be reasonable to split out `libgfortran` as it isn't really a required C/C++ library.

asmeurer:
gfortran is a very common dependency (in the case of OS X, it's the one library that isn't already part of the system). I think it's valuable to leave it in here.

ccordoba12:
@asmeurer, should we close this one then?

jakirkham:
Yeah, I'll open two new ones.

PR #319, opened 2015-06-03
Adds a recipe for PyMongo 3.0.3

jakirkham:
Closes conda/conda-recipes#345. Moves the old recipe to `pymongo2` (for those who still want that version) and updates the recipe for PyMongo to 3.0.3 in the folder. Tested on Mac (10.9) and Linux (CentOS 6.3).


jakirkham:
Should add that I have only renamed the folders in this repo. The names of packages are **exactly** the same in the `meta.yml`.

groutr:
I think pymongo3 should remain as pymongo (and pymongo should always track the latest stable release).

jakirkham:
> I think pymongo3 should remain as pymongo (and pymongo should always track the latest stable release).

@groutr, I agree. Changed to have one copy of PyMongo, which is version 3.

Was just being cautious in case someone thought we should do it differently.

groutr:
I probably should have also stated that I'm fine with having a pymongo2 package for those that need it.  I just think pymongo should always track the latest stable release.

jakirkham:
Done.

jakirkham:
This now depends explicitly on pymongo 2 updates occurring first ( https://github.com/conda/conda-recipes/pull/345 ).

jakirkham:
Updated to 3.0.3 and retested on Mac (10.9) and Linux (CentOS 6.3).

jakirkham:
@groutr, I think this is ready to go if you have no more comments.

jakirkham:
@groutr, please let me know if you have any more thoughts on this. Happy to make changes as needed.

jakirkham:
Can we merge this please?

jakirkham:
Noticed there was an issue with pymongo and iPython. ( https://github.com/ipython/ipython/issues/8657 ) This has been resolved. A fix is already released for iPython 3.x in rel-3.2.2. A fix for Jupyter will be available as soon as the next ipyparallel release occurs.

jakirkham:
Made note requesting a new release of iPython 3.x. ( https://github.com/ContinuumIO/anaconda-issues/issues/457 )

jakirkham:
@groutr, would you be willing to take another look at this please?

msarahan:
I disagree with Ryan on keeping the pymongo package up-to-date with the latest major version API.  Ideally, I think we really need to include the major version in all package names.  Updating this can cause serious backwards compatibility breakage.  The proposed behavior here is different from what we currently plan on doing with Qt (4->5 soon), I think.

This recipe otherwise looks OK.  I personally would clean up more of the comments left from the template recipe generation, but they're not hurting anything.

jakirkham:
@msarahan, alright, how does this look?

jakirkham:
Actually, this probably should explicitly include `mongodb`, no? Let me add that real quick.

msarahan:
Yes, I like this better.  Adding mongodb to meta.yaml as build and/or runtime requirement would be good.  I don't know if build is required, but I trust that you do.

I will confer with Ryan today to make sure we have a solid plan going forward on version numbers in package names, then we will merge this.

jakirkham:
> Yes, I like this better.

Cool.

> Adding mongodb to meta.yaml as build and/or runtime requirement would be good.

Done.

> I don't know if build is required, but I trust that you do.

It could probably be dropped. I feel a little safer with it there.

> I will confer with Ryan today to make sure we have a solid plan going forward on version numbers in package names, then we will merge this.

Sounds good. Please keep me posted.

msarahan:
I have discussed this with @ilanschnell and @groutr .  I was wrong.  It absolutely makes sense to have a single, always-up-to-date package.  However, there are things that go with this:

  - consumers of any package would do well to pin versions: http://conda.pydata.org/docs/faq.html#pinning-packages
  - it is probably advantageous to keep older recipe FOLDERS around when updating to a new major version.  This might make it simpler to pull in bugfix versions until the older major version is truly deprecated.

It is a bad idea to include the major revision in the package name, because this is effectively defeating the whole point of version specification in any package manager.

Sorry to ask you to change this again, but could you please:
- make the new version have no major version in the folder name
- make the old version have major version 2 in the folder name
- keep the names of BOTH packages as pymongo 

    package: 
        name: pymongo

Thanks.  I'm going to go hunt for docs on this topic, and write them if necessary.

jakirkham:
No worries. I had the old branch hanging around from before so it wasn't to bad to grab the old commit.

I think I only changed the folder name not the package name, no? If I messed that up, feel free to remind me tomorrow.

Otherwise this should be in working order now.

msarahan:
You did not mess this up, I just wanted to document what I concluded was the "Right Way" to do it.

This looks good.  I am merging.

jakirkham:
That was fast. Thanks.

>  I just wanted to document what I concluded was the "Right Way" to do it.

Sounds perfectly reasonable. Also, thanks for looking into docs.

I wonder whether the README here should be expanded on. Adding a docs folder might confusing consider the number of packages. Alternatively, a contributing or other type document could be added to clarify this procedure.

PR #320, opened 2015-06-03
fix test-noarch

hargup:
PR conda/conda-build#317 introduced major changes in the noarch option which made earlier test example obsolete. This test package is taken from conda-build/example_packages/noarch_python


hargup:
@asmeurer

PR #321, opened 2015-06-10
Updated s3cmd to 1.5 and added supporting library python-magic

bkreider:
Updated s3cmd to version 1.5.  It was on a beta release.  Also added python-magic because it was a supporting library.  

Note: i needed to patch setup.py because distutils tries to download packages.  My patch removes the required packages, because the dependency is expressed in the meta.yaml.

### DIFF
--- a/python-magic/meta.yaml
+++ b/python-magic/meta.yaml
@@ -1,0 +1,59 @@
+package:
+  name: python-magic
+  version: !!str 0.4.6
+
+source:
+  fn: python-magic-0.4.6.tar.gz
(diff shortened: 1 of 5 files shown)


python-magic/meta.yaml
```diff
@@ -0,0 +1,59 @@
+package:
+  name: python-magic
+  version: !!str 0.4.6
```
damianavila: Why this version with the `!!`? Probably I am missing something in the yaml specification capabilities... or something else... so asking you to learn a little bit...

msarahan: @bkreider any answer for @damianavila ? I'd also like to learn.

PR #325, opened 2015-06-12
Add jq, a sed-like command line tool for JSON

jakirkham:
Tested on Mac OS 10.9 and CentOS 6.3.

### DIFF
--- a/jq/build.sh
+++ b/jq/build.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+chmod +x configure
+
+./configure --prefix=$PREFIX 
+
+make -j
+make install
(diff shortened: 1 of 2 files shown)


jakirkham:
Thanks @ccordoba12. Sorry about this, but I have bumped the version on this recipe, which is in a new pull request ( https://github.com/conda/conda-recipes/pull/326 ).

jakirkham:
Just for anyone, who looks at this in the future. A really useful case of `jq` is parsing iPython notebooks. An example is provided here of how one could use this as a `jq`-based filter in `git` to knock out all output cells ( https://gist.github.com/takluyver/bc8f3275c7d34abb68bf#comment-1416093 ).

PR #326, opened 2015-06-12
Update jq to 1.4

jakirkham:
Updated the version of `jq`. Originally, this included `bison`. This has been dropped as patching `configure` allowed this error to be bypassed. As it was determined that `bison` was not being used, excluding this dependencies did not cause a problem.

### DIFF
--- a/jq/configure.patch
+++ b/jq/configure.patch
@@ -1,0 +1,18 @@
+diff --git configure configure
+index 44eee99..a5a18d3 100755
+--- configure
++++ configure
+@@ -12958,9 +12958,10 @@ elif $YACC --help | grep -- --warnings >/dev/null; then
+   { $as_echo "$as_me:${as_lineno-$LINENO}: result: ok" >&5
+ $as_echo "ok" >&6; }
+ else
+-  { $as_echo "$as_me:${as_lineno-$LINENO}: result: no" >&5
+-$as_echo "no" >&6; }
+-  as_fn_error $? "You need bison version 2.5 or more recent." "$LINENO" 5
++  :
++#  { $as_echo "$as_me:${as_lineno-$LINENO}: result: no" >&5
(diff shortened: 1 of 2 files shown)


ccordoba12:
You also need `libtool` to build `jq`, at least according to its Readme

https://github.com/stedolan/jq/blob/master/README.md

ccordoba12:
And it can also be built without `bison`, if you are not planning to modify it.

jakirkham:
> And it can also be built without bison, if you are not planning to modify it.

Unfortunately, that feature is not implemented until version 1.5 ( https://github.com/stedolan/jq/issues/689#issuecomment-74446461 ), which has not yet been released.

The only reason I included `bison` in this PR as opposed to the last one is the version from Apple's SDK is too old in this case and it fails to build if it doesn't find an acceptable version. On Linux, it looked for `bison`, which I didn't have, and proceeded on without it.

<br>
> You also need libtool to build jq, at least according to its Readme
> 
> https://github.com/stedolan/jq/blob/master/README.md

As far as `libtool` goes, on Mac the version included worked out of the box. On Linux, this was a bit weird, it appeared to somehow create `libtool` in the `jq` work directory, which I don't understand.

jakirkham:
I was able to patch the configure file based on this suggestion ( https://github.com/stedolan/jq/issues/689#issue-56239189 ). As a consequence, I have changed this PR to ignore the bison constraint. I created a separate PR for bison as it seems useful to have regardless ( https://github.com/conda/conda-recipes/pull/329 ). This has been retested and works on both OSes.

jakirkham:
@ccordoba12, I think I have addressed your concerns. This now works without `bison` and no longer includes the recipe. I have updated the summary to reflect this. Please let me know if you have any more suggestions.

ccordoba12:
Thanks.

PR #327, opened 2015-06-12
WIP: JDK build fixes

jakirkham:
It appears that there are some issues with the JDK recipe now. In particular, the download strategy doesn't work due to cookies and redirects. I have fixed this problem.

However, it seems there are other issues. For instance, at least one of the libraries on Mac fails to be amended by `install_name_tool`. On Linux, there are libraries that appear to missing.

As a consequence, it may be necessary to switch to building OpenJDK from source or something similar.

### DIFF
--- a/jdk/build.sh
+++ b/jdk/build.sh
@@ -3,7 +3,7 @@
 #### Download in build because we have to set cookies ####
 # Check if we're on OS X
 if [ $(uname) = "Darwin" ]; then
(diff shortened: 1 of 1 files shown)


jakirkham:
@ilanschnell, just so you are aware. This doesn't build for me; however, it wouldn't even download for me before. So, this is an improvement. The issues with "building" JDK this way are issues that the current strategy probably can't resolve.

ilanschnell:
I was merging PRs last night, and must have overlooked "WIP".  Well, at least your JDK recipe has more visibility now, that is it in master.

jakirkham:
Not a problem. Just wanted to make sure there was no confusion about its state.

PR #328, opened 2015-06-13
WIP: Add awscli (and required dependencies)

jakirkham:
Allows for the AWS command line interface to built/installed within `conda`.

At present, this still does not work because `awscli` fails to properly install the `aws` script in the `bin` directory, which result in something that looks the text below. This problem does not occur when using `pip` with the package. I will need to investigate further to determine the cause.

    #!/opt/anaconda1anaconda2anaconda3/bin/python
    # EASY-INSTALL-SCRIPT: 'awscli==1.7.34','aws'
    __requires__ = 'awscli==1.7.34'
    __import__('pkg_resources').run_script('awscli==1.7.34', 'was')

### DIFF
--- a/awscli/bld.bat
+++ b/awscli/bld.bat
@@ -1,0 +1,8 @@
+"%PYTHON%" setup.py install
+if errorlevel 1 exit 1
+
+:: Add more build steps here, if they are necessary.
+
+:: See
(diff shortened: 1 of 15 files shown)


ilanschnell:
Thanks

jakirkham:
So, I think this is still broken. Just FYI.

PR #329, opened 2015-06-13
Add bison

jakirkham:
Borrowed bison from @jdreaver and relaxed the Linux constraint. Tested on Mac OS 10.9 and CentOS 6.3.


jakirkham:
@ccordoba12 Here is a separate PR for `bison` as it still seems like a valuable recipe to keep even though it is no longer needed to build `jq`.

ccordoba12:
Thanks.

PR #330, opened 2015-06-15
WIP: Build and install go

jakirkham:
Installs such that `$GOROOT`=`$PREFIX`. This works and has a test `go` program to verify everything is working once `go` is built. This has been tested on Mac OS 10.9 and CentOS 6.3.

### DIFF
--- a/go/build.sh
+++ b/go/build.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+# Tell go where it will be installed is not the same as where it is being built.
+export GOROOT_FINAL=$PREFIX
+
+# Change to `src` build and then move out.
+cd src
+./all.bash
+cd ..
+
+# Install everything manually.
+# Note go expects to find `src` for some operations so it **must** be installed.
+test -d bin && mkdir -p $PREFIX/bin && mv bin/* $PREFIX/bin
+test -d include && mkdir -p $PREFIX/include && mv include/* $PREFIX/include
+test -d lib && mkdir -p $PREFIX/lib && mv lib/* $PREFIX/lib
+test -d pkg && mkdir -p $PREFIX/pkg && mv pkg/* $PREFIX/pkg
+test -d sbin && mkdir -p $PREFIX/sbin && mv sbin/* $PREFIX/sbin
+test -d slib && mkdir -p $PREFIX/slib && mv slib/* $PREFIX/slib
+test -d src && mkdir -p $PREFIX/src && mv src/* $PREFIX/src
+rm -rf $PREFIX/doc/go && mkdir -p $PREFIX/doc/go && mv doc/* $PREFIX/doc/go
--- a/go/build.sh
+++ b/go/build.sh
@@ -1,20 +1,14 @@
 #!/bin/bash
 
-# Tell go where it will be installed is not the same as where it is being built.
-export GOROOT_FINAL=$PREFIX
+# Move the go directory into $PREFIX/ so it can be built in its install location.
+cd .. && mv go $PREFIX/ && cd $PREFIX/go
 
 # Change to `src` build and then move out.
 cd src
 ./all.bash
 cd ..
 
-# Install everything manually.
-# Note go expects to find `src` for some operations so it **must** be installed.
-test -d bin && mkdir -p $PREFIX/bin && mv bin/* $PREFIX/bin
-test -d include && mkdir -p $PREFIX/include && mv include/* $PREFIX/include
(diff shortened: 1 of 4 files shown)


jakirkham:
Unfortunately, the end user must set `$GOROOT` as is specified under topic "Installing to a custom location" ( http://golang.org/doc/install ). 

If someone has a suggestion to alleviate this problem, I am open to suggestions.

jakirkham:
So, I have fixed the `$GOROOT` problem in the second commit. However, this requires a post-link step, which is somewhat intensive. Also, relocation must explicitly be disallowed. This does test out on the same flavors of Mac and Linux.

jakirkham:
Sped up post-link step by skipping tests.

ccordoba12:
What do you mean by

>  relocation must explicitly be disallowed

Does that mean that your Go package can't be installed in a conda environment?

go/build.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# Move the go directory into $PREFIX/ so it can be built in its install location.
+cd .. && mv go $PREFIX/ && cd $PREFIX/go
```
ccordoba12: This is not the way conda-build works. Doesn't Go has a `make install` step or something?

go/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+  binary_relocation: false
+
+test:
+  files:
+    - hello.go
+
+  commands:
+    - go help
+    # Compile and run a test program. Need to set GOROOT explicitly see topic "Installing to a custom location". ( http://golang.org/doc/install )
```
ccordoba12: Please break this line in two or three. It's too long.

go/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - go run hello.go
+
+about:
+  home: http://golang.org
+  license: BSD
+  summary: Go is an open source programming language that makes it easy to build simple, reliable, and efficient software.
```
ccordoba12: Also break this line (if possible).

go/build.sh
jakirkham: Unfortunately not. It wants to be installed in place. I tried other ways of going about it, but this ended up being the simplest and is what they recommend. See "Installing to a custom location" ( https://golang.org/doc/install ).

go/meta.yaml
jakirkham: Yeah, that comment is outdated so we can cut it.

go/meta.yaml
jakirkham: Ok, see what I can do.

jakirkham:
> What do you mean by

>> relocation must explicitly be disallowed

> Does that mean that your Go package can't be installed in a conda environment?

I mean that we have set this `binary_relocation: false`.

jakirkham:
@ccordoba12, at present, even excluding your suggestions, I'm not sure if this is ready to go (no pun intended).

I find that we require some recompilation as a post-link step (to avoid the `GOROOT` problem). This may mean we need to include a compiler at runtime or something like this and hence may be hinged on this ( https://github.com/conda/conda-recipes/pull/279 ) being accepted. Personally, I would rather avoid this post-link step if possible, but have had no luck so far.

I have tried to find ways around this, but somehow `go` is hanging on to the old path somewhere. The only objects that still appear to have the old path are static libraries that `go` builds. Maybe these are the problem? Do you know if `conda` handles these in any way?

Alternatively, I could look into excluding the static libraries, if possible.

go/build.sh
jakirkham: Take a look at my first commit if you want to see another way, in which I tried to install it into `$PREFIX`. From what I can tell, it seemed to work, but I have no idea what might break as it is non-standard.

msarahan:
@jakirkham what's the status on this? Is it ready for us to look at?

jakirkham:
I'll have to look at this again. Is there a good way to set an environment variable when any time an environment is activated? For instance, what I would like to do here is set `GOROOT`. ( https://github.com/golang/go/blob/master/README.md#binary-distribution-notes ) Previously, I was doing a rebuild of everything, but that is very burdensome for package installation and I think setting this environment variable is just easier to deal with.

msarahan:
See http://stackoverflow.com/q/31598963/1170370

jakirkham:
Thanks @msarahan. I remember seeing this somewhere before.

Is there a way for this to be attached to a package like the `go` package or is this just environment wide? I suppose I could patch those files, but using pre-link and post-unlink steps, but it would be nice if there was a cleaner way to do this.

PR #331, opened 2015-06-16
Openblas package with numpy and scipy linked to it.

abergeron:
This uses the same feature that mkl does so that you can say `conda install openblas numpy`and it will select the `[openblas]` version of numpy (or scipy).  I've tried getting the automatic reinstall to work (just like it does for mkl) but that seems to require something else.

In the meantime this could be useful for people that want to have a faster blas that ATLAS but not use mkl.  This is also useful for Theano which requires a dynamic library for the blas.

I've also added a comment in the meta file for numpy explaining the extra step required to make this work that isn't documented anywhere.


abergeron:
These packages are also available in binary form for linux-64 on my binstar page: https://binstar.org/abergeron

openblas-libs/build.sh
```diff
@@ -0,0 +1,5 @@
+#/bin/bash
+
+make USE_THREAD=0 DYNAMIC_ARCH=1 NO_STATIC=1 PREFIX=$PREFIX
```
jakirkham: Instead of setting `USE_THREAD=0`, could you set `NO_AFFINITY=1 NUM_THREADS=1`. This will default to 1 thread for all practical purposes, but allow people to override if they wish by setting the environment variable `OPENBLAS_NUM_THREADS` or one of the other ones if they are relevant ( <https://github.com/xianyi/OpenBLAS#set-the-number-of-threads-with-environment-variables> ) before loading the library. The `NO_AFFINITY` is necessary to avoid issues when forking. I have also found this works well on certain flavors of Linux that otherwise run into trouble. Also, it avoids an issue with building R ( https://stat.ethz.ch/pipermail/r-sig-hpc/2012-April/001348.html ).

abergeron:
I've made the suggested change to the OpenBLAS build script.

jakirkham:
Thanks.

jakirkham:
Also, `openblas` and `openblas-lib` can be merged into the same recipe. It is perfectly acceptable to use the `track_features` parameter in a non-metapackage recipe.

jakirkham:
I have some changes for you that I have worked on. Once they are complete, I would like to PR them to you against your branch here.

ilanschnell:
This is fantastic.  Is this ready to be merged?

jakirkham:
Not yet, I am still seeing some build problems. Plus, it requires some more changes.

PR #335, opened 2015-07-01
Imageio v1.3

almarklein:
New release for imageio. Also changed that the freeimage lib is now packed into the library, so that it won't have to be auto-downloaded on first use.

Any chance that CA can build this conda package? I'd love to say that ppl can do `conda install imageio` to install it.

### DIFF
--- a/imageio/bld.bat
+++ b/imageio/bld.bat
@@ -1,4 +1,4 @@
-"%PYTHON%" setup.py install
+"%PYTHON%" setup.py build_with_fi install
 if errorlevel 1 exit 1
 
 :: Add more build steps here, if they are necessary.
--- a/imageio/meta.yaml
+++ b/imageio/meta.yaml
@@ -1,22 +1,22 @@
 package:
   name: imageio
-  version: 1.2
+  version: 1.3
 
 source:
-  fn: imageio-1.2-linux32.zip  [linux32]
-  url: https://pypi.python.org/packages/source/i/imageio/imageio-1.2-linux32.zip  [linux32]
+  fn: imageio-1.3-linux32.zip  [linux32]
+  url: https://pypi.python.org/packages/source/i/imageio/imageio-1.3-linux32.zip  [linux32]
   
-  fn: imageio-1.2-linux64.zip  [linux64]
-  url: https://pypi.python.org/packages/source/i/imageio/imageio-1.2-linux64.zip  [linux64]
+  fn: imageio-1.3-linux64.zip  [linux64]
+  url: https://pypi.python.org/packages/source/i/imageio/imageio-1.3-linux64.zip  [linux64]
   
-  fn: imageio-1.2-osx64.zip  [osx]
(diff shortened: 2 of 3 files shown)


almarklein:
cc @ccordoba12 and @blink1073

So maybe we should *not* include the freeimage lib in the package, but have a separate conda package for freeimage and depend on that?

blink1073:
For conda purposes, I think it makes sense to have libfreeimage be a separate dependency for imageio.

https://github.com/conda/conda-recipes/tree/master/freeimage

ccordoba12:
I agree with @blink1073, it should managed as a dependency of `imageio`

imageio/bld.bat
```diff
@@ -1,4 +1,4 @@
-"%PYTHON%" setup.py install
+"%PYTHON%" setup.py build_with_fi install
```
ccordoba12: What does this `build_with_fi` flag mean?

almarklein: It means that the freeimage lib is downloaded and included in the package.

almarklein:
Cool, in that case hold of the merging until we change it back, And we need to make sure that imageio can find freeimage where its installed in the conda env.

imageio/meta.yaml
```diff
@@ -1,26 +1,27 @@
 package:
   name: imageio
-  version: 1.2
+  version: 1.3
 
 source:
-  fn: imageio-1.2-linux32.zip  [linux32]
-  url: https://pypi.python.org/packages/source/i/imageio/imageio-1.2-linux32.zip  [linux32]
+  fn: imageio-1.3-linux32.zip  [linux32]
+  url: https://pypi.python.org/packages/source/i/imageio/imageio-1.3-linux32.zip  [linux32]
```
ccordoba12: Why do you need a *source* file per platform and arch? i don't get it ;-)

almarklein: Oh wait. Now I'm confused. These per-platform source files include the freeimage lib .... it should be enough to *either* to this or use the `build_with_fi` flag.

Or none of the above, really, if anaconda has freeimage :P

ccordoba12: Yep, none of the above if Anaconda comes with freeimage as an independent library :-)

Well, maybe we can leave these source files for Windows, but definitely not for Mac and Linux.

almarklein:
When can we `conda install freeimage`? When we do, I can test/make sure that imageio can find the lib. Then do a new release and update this recipe...

ccordoba12:
It depends on Ilan, but right now you can use the recipe mentioned by @blink1073 above to build a conda package for `freeimage` on Linux/Mac and then add it as a dependency of `imageio`.

That way you can test if conda-build works correctly in this case (although i don't expect you to have big problems with that :-).

almarklein:
Ah yes, I can also build it myself :P (It must be getting late...)

almarklein:
Closing this in favour of a new PR to come.

PR #342, opened 2015-07-08
Build with littlecms2 support

bgreen-litl:
Pillow has an optional module PIL.ImageCms which is only available when
compiled with littlecms2.

Fixes https://github.com/ContinuumIO/anaconda-issues/issues/371


ilanschnell:
I agree, that's useful.  Thanks

PR #345, opened 2015-07-15
pymongo: Update to version 2.8.

jakirkham:


jakirkham:
@groutr, maybe we should do this one before adding pymongo version 3?

jakirkham:
Nevermind, the other PR now contains this change first and will close out this PR, as well.

PR #360, opened 2015-07-27
update swig

jschueller:
v3.0.6 has a lot of python fixes including a regression fix from 3.0.2:
https://github.com/swig/swig/blob/master/CHANGES


ilanschnell:
Thanks for updating swig!

Issue #351, opened 2015-07-20
Improve Freeimage recipe


almarklein:
In #335 we agreed that the imageio package should depend on the freeimage conda package, rather than shipping its own freeimage lib. I tested that imageio can find the lib where it is put inside the conda env.

However, the current freeimage recipe has a few problems:

* It is outdated (the version of FI that it builds is rather old)
* It only build 64 libs (fails on 32 bit systems), because the recipe sets some compiler flags, which should not be necessary. Building FI should just be `make\nmake install`.
* Not sure if it builds on OSX, see also #141

almarklein:
I'd be happy to work on a better version. However, my Mac VM is broken so I cannot test for OS X.

ccordoba12:
@almarklein, I can help you with OS X, so please send a pull request with your improvements :-)

almarklein:
Ah, thanks Carlos. Great, I'll get on it when I find some time.

PR #362, opened 2015-07-27
Freeimage to 3.17, add test, works on all platforms (osx needs testing)

almarklein:
Fixes issue #351

Tested on:
* Linux 64 bit
* Linux 32 bit
* Windows XP 32 bit
* Windows 7 64 bit
* todo: osx

cc @ccordoba12


almarklein:
Compiling on Windows is apparently a pain, and not something that I would call my expertise. The recipe now simply copies in the precompiled libs.

Needs testing on OSX, you may need to modify some paths to copy the libs from.

jakirkham:
So, I have tried building 3.17 on Mac in the past and retried it again today using your changes. Unfortunately, this does not work.

The last time I tried to do this, I gave up and simply tried to get the exist `freeimage` recipe to work. This was still a significant task and require a fair amount of research, but it builds through and I have it here ( https://github.com/jakirkham/conda-recipes/tree/fix_freeimage_2 ). Here's a synopsis of the changes.

* `sed` in-place requires an extension be specified for the original files. Will still work on Linux.
* `sed` gets hung up on what it considers binary files. Must convince it that it can work through them.
* One header file is missing an include.
* A C++ source file does not appropriately cast types.
* The OS X Makefile is completely outdated. So it needed to be modernized
    * Ancient `gcc` specified `4.0`. Just use `gcc` with any version.
    * Build steps for ancient platforms namely PPC and 32-bit Mac. Drop from the Makefile.

jakirkham:
FWIW, I went ahead and rebased the changes from that branch onto yours. Also, I updated the patches to reflect the current state of the code. The result is in this branch ( https://github.com/jakirkham/conda-recipes/tree/fix_freeimage-3.17 ). However, this runs into a different error. See the traceback below.

    In file included from Source/LibJXR/./image/decode/decode.c:40:
    Source/LibJXR/image/sys/strcodec.h:87:25: error: expected ';' after top level declarator
    typedef unsigned __int64 U64;
                            ^
                            ;
    Source/LibJXR/image/sys/strcodec.h:95:18: error: use of undeclared identifier 'UINTPTR_T'
    CT_ASSERT(sizeof(UINTPTR_T) == sizeof(void*), strcodec1);
                     ^
    Source/LibJXR/image/sys/windowsmediaphoto.h:249:64: note: expanded from macro 'CT_ASSERT'
    #define CT_ASSERT(exp, uniq) typedef char __CT_ASSERT__##uniq[(exp) ? 1 : -1] // Caller must provide a unique tag, or this fails to compile under GCC
                                                                   ^
    In file included from Source/LibJXR/./image/decode/decode.c:40:
    Source/LibJXR/image/sys/strcodec.h:96:18: error: use of undeclared identifier 'INTPTR_T'
    CT_ASSERT(sizeof(INTPTR_T) == sizeof(void*), strcodec2);
                     ^
    Source/LibJXR/image/sys/windowsmediaphoto.h:249:64: note: expanded from macro 'CT_ASSERT'
    #define CT_ASSERT(exp, uniq) typedef char __CT_ASSERT__##uniq[(exp) ? 1 : -1] // Caller must provide a unique tag, or this fails to compile under GCC
                                                                   ^
    In file included from Source/LibJXR/./image/decode/decode.c:40:
    Source/LibJXR/image/sys/strcodec.h:130:11: error: use of undeclared identifier 'PACKETLENGTH'
        U8 P0[PACKETLENGTH];        // packet circular buffer 0
              ^
    Source/LibJXR/image/sys/strcodec.h:131:11: error: use of undeclared identifier 'PACKETLENGTH'
        U8 P1[PACKETLENGTH];        // packet circular buffer 1
              ^
    Source/LibJXR/image/sys/strcodec.h:135:15: error: use of undeclared identifier 'PACKETLENGTH'
            U8 P2[PACKETLENGTH];
                  ^
    Source/LibJXR/image/sys/strcodec.h:158:11: error: use of undeclared identifier 'PACKETLENGTH'
        U8 P3[PACKETLENGTH];    // index packet buffer
              ^
    7 errors generated.
    make[1]: *** [Source/LibJXR/./image/decode/decode.o] Error 1
    make: *** [default] Error 2
    + exit 1
    Command failed: /bin/bash -x -e /Users/kirkhamj/Developer/Conda/conda-recipes/freeimage/build.sh

jakirkham:
Alright, I have fixed several build errors and now have a working Mac copy. Let me clean it up a bit and I will PR it against your branch.

jakirkham:
As a side note, it might be easier to work with MinGW if `make` were present. ( https://github.com/ContinuumIO/anaconda-issues/issues/384 )

jakirkham:
@almarklein, I have placed my changes in this PR ( https://github.com/almarklein/conda-recipes/pull/1 ).

almarklein:
Thanks @jakirkham. There's quite a few patches in there. Would it be an idea to contribute these to the FreeImage project instead? I know FI uses sourceforge and SVN (yuk), but that we would not have to verify that our patches work every time we update. Also see [this discussion](http://sourceforge.net/p/freeimage/discussion/36111/thread/f2fae857/) about (not) moving FI to GH.

> As a side note, it might be easier to work with MinGW if make were present.

And it would feel safer; I feel slightly uncomfortable with downloading binaries from sourceforge.

jakirkham:
> Thanks @jakirkham. There's quite a few patches in there. Would it be an idea to contribute these to the FreeImage project instead? I know FI uses sourceforge and SVN (yuk), but that we would not have to verify that our patches work every time we update. Also see this discussion about (not) moving FI to GH.

@almarklein, I'd like to have this sort of discussion on the PR itself. So, I will respond to it over there.

> And it would feel safer; I feel slightly uncomfortable with downloading binaries from sourceforge.

Agreed.

ccordoba12:
I'm glad @jakirkham has helped to solve all compilation issues on Mac. That's really great! Thanks for that.

Please don't consider using MinGW on Windows (at least not the version that comes with Anaconda). It's very unreliable, specially on Win 64.

There is a pending PR in conda-build to support Visual Studio. Once that's merged, people can start using it to create Windows packages the right way :-) (By the way, we use Visual Studio internally for all our packages :-).

ccordoba12:
I'll try to merge this tomorrow, after a bit of testing.

jakirkham:
> I'm glad @jakirkham has helped to solve all compilation issues on Mac. That's really great! Thanks for that.

Of course, I am very eager to see a working copy of FreeImage.

> Please don't consider using MinGW on Windows (at least not the version that comes with Anaconda). It's very unreliable, specially on Win 64.
>
> There is a pending PR in conda-build to support Visual Studio. Once that's merged, people can start using it to create Windows packages the right way :-) (By the way, we use Visual Studio internally for all our packages :-).

Ah, thanks for sharing. I will have to take a look.

> I'll try to merge this tomorrow, after a bit of testing.

Of course, please let me know if you run into any issues on the Mac/Linux side.

jakirkham:
@ccordoba12, have you had a chance to play with this at all?

ccordoba12:
Nop, sorry. Will do it this afternoon :-)

jakirkham:
Not a problem, just wanted to touch base.

almarklein:
I have an OSX Snow Leopard VM now. I tested and it works fine. Also tested that imageio picks it up and can read images with it.

The only thing is that I needed to revert to `gcc` and `g++`, instead of clang, because for some reason clang would not build.

@ccordoba12 what machine is used by CA to build packages for OSX? I recall it's Leopard?

jakirkham:
For some reason, I had thought that I had build issues using `gcc/g++`, but I just tried building with those compilers now and everything checked out thus far.

@almarklein, regarding these additional tests that you ran, is there code for them somewhere. Also, since you have a Snow Leopard machine with you, I was curious what `cc` and `c++` point to. Are they the GNU compilers or are they Clang compilers? Finally, I believe I know the answer to this question, but wanted to check, are these GNU compilers actually GNU compilers or are they frontends to LLVM?

almarklein:
@jakirkham no, these test consist of me doing `conda install freeimage` then loading an image with imageio, and verifying that it uses conda's freeimage.

almarklein:
Do you mean this?

    $ which gcc
    /usr/bin/gcc
    $ which g++
    /usr/bin/g++

jakirkham:
@almarklein, on the testing side, ok. Could you please share the image (if possible) and steps? I just want to be sure that the `gcc`/`g++` build on my machine is really working.

Regarding compiler stuff, sorry, should have been clearer, I was meaning this below. These, of course, are for my machine.

    $ cc --version
    Apple LLVM version 5.0 (clang-500.2.79) (based on LLVM 3.3svn)
    Target: x86_64-apple-darwin13.4.0
    Thread model: posix
    $ c++ --version
    Apple LLVM version 5.0 (clang-500.2.79) (based on LLVM 3.3svn)
    Target: x86_64-apple-darwin13.4.0
    Thread model: posix
    $ gcc --version
    Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
    Apple LLVM version 5.0 (clang-500.2.79) (based on LLVM 3.3svn)
    Target: x86_64-apple-darwin13.4.0
    Thread model: posix
    $ g++ --version
    Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
    Apple LLVM version 5.0 (clang-500.2.79) (based on LLVM 3.3svn)
    Target: x86_64-apple-darwin13.4.0
    Thread model: posix

almarklein:
@jakirkham I just tested that the result of `imageio.imread('chelsea.png')` was indeed a numpy array of expected shape. To verify that imageio was using my new freeimage I checked that it did not auto-download its own freeimage lib (and that it did when I tried again after conda removing freeimage). Not a very straightforward test :P  

I wanted to propose that you upload your lib and let me give a try, but I don't think it will work, since my OS is of a lower version.

Also: \i686-apple-darwin10-gcc-4.2.1 (GCC) 4.2.1 (Apple Inc. build 5646)

almarklein:
ping

@jakirkham are you happy?
@ccordoba12 you want to give this a go?

jakirkham:
@almarklein, pretty much, did you already switch the patch to use `gcc`/`g++`?

almarklein:
Oef, good call! I forgot to push that. I did now.

ccordoba12:
Will do this weekend. I'm busy right now with a client's project :-)

freeimage/bld.bat
```diff
@@ -0,0 +1,12 @@
...
+  set BDIR=Dist\x64
+) else (
+  set BDIR=Dist\x32
+)
+
+copy %BDIR%\FreeImage.dll %LIBRARY_BIN%\FreeImage-3.17.0.dll
+copy %BDIR%\FreeImage.dll %LIBRARY_BIN%\FreeImage.dll
+copy %BDIR%\FreeImage.h %LIBRARY_INC%\FreeImage.h
+copy %BDIR%\FreeImage.lib %LIBRARY_LIB%\FreeImage.lib
+copy C:\Windows\System32\vcomp120.dll %LIBRARY_BIN%\vcomp120.dll || exit "needs VC redist from http://www.microsoft.com/en-us/download/details.aspx?id=40784"
```
patricksnape: :disappointed: You shouldn't do that! Please see Steve Dower's [excellent discussion on why you shouldn't mix Vistual Studio CRTs](http://stevedower.id.au/blog/building-for-python-3-5/). Luckily, we can solve this problem - have a look at my [conda recipe for freeimage](http://www.github.com/menpo/conda-freeimage).

jakirkham:
@patricksnape, if you want to contribute a fix, I'm sure @almarklein would appreciate it. After all, that's how Mac support was added by yours truly.

patricksnape:
@jakirkham Without trying to sound like I'm dragging my feet here, what is the point of this repository? Because most of the recipes are out of date or don't work at all - and those that do work are mostly only for some subset of platforms (mostly Unix only). And since Anaconda is built using its own proprietary system whereby not every package in Anaconda is in this repo - what is this repo for? I've got recipes for quite a lot of popular packages, but I just put them on my own channel where I know I will actively maintain them for all systems.

jakirkham:
@patricksnape, below, I have tried to write a thoughtful and considerate response.
<br>

> In open source, we feel strongly that to really do something well, you have to get a lot of people involved. - Linus Torvalds [1]

I quote this not to try and be a pain, but only to point out a very important feature of open source software. Namely, that open source excels where other strategies fail by getting more eyes on the code and more hands involved. Now I don't say this to be mean, but there are simply more people looking at the code here and helping maintain it. The fact that it is maintained and reviewed by the people at Continuum gives it some level of authority that anyone elses' repo just won't have.

As for "out of date", they do periodically sync these recipes with a closed source set of recipes. Its probably due for another sync if that's your point.

Regarding, your point that some "don't work at all". If you find ones that don't work, you are free to do the same things you can on any other repo. Open issues and PRs. However, complaining on this PR that there is some configuration that they haven't tried that doesn't work seems unfair. Especially, when a fair number of configurations have been tried and you haven't provided details regarding the particular build issues you are experiencing.

Regarding the "mostly Unix" point, there are many challenges when building for Windows. Not the least of which is the highly varied environments that packages are built in. For instance, some packages require some level of Unix-like features. Should they be built in a MinGW environment? How should that environment be packaged by conda, if at all. In most cases, the best that can be done is to simply copy libraries from prebuilt packages in to the environment. Another major concern is access to a Windows environment for testing. Even when this is satisfied, there are enough differences between OS versions to cause difficulties. This combined with a generally slower uptake of OS releases amongst Windows users makes it increasingly difficult to find a working strategy. As you have nicely pointed out above, sometimes even copying over libraries may not be as simple as it seems.

That all said, and please don't take this the wrong way. If you think this work is valueless, I am a bit confused as to why you decided to comment here at all.

<br>
[1] - http://www.nytimes.com/2003/09/28/magazine/the-way-we-live-now-9-28-03-questions-for-linus-torvalds-the-sharer.html

ccordoba12:
@patricksnape, about Windows: It's been very hard for us to find developers with the expertise to build complex packages on Windows. Fortunately, there are several people now on the Anaconda team that are helping with that.

patricksnape:
@jakirkham Reread my comment and admit that it was rude and apologise. We got off on the wrong foot here and I phrased that question horribly. In short, what I meant to ask was:

"~~Is this repository actively maintained?~~ (You answered this) Should I be contributing all of my recipes here - do they all get synced back in to conda mainline? Were I to add recipes for projects, would they end up being installable via the main anaconda channel? Also, as an aside, I've noticed that the manner in which this repository is suboptimal with regards to checking for correctness since it would be great if automated builds were kicked off for each recipe rather than relying on manual checking by developers. I have put a lot of work into maintaining automated cross-platform builds for many popular packages and I would hate to see that go to waste by contributing code here which needs manual verification."

I commented across here because I have written a working freeimage build on all platforms in the link I posted previously. And I thought you might want those scripts. Since I'm obviously (sincerely) in the wrong here, let me introduce myself as someone with quite a lot of experience with fighting conda-build, Windows, OSX and Linux to provide automated builds. I have a keen interested in imageio and so I'm happy to put a PR up for my freeimage and imageio scripts - but they will clash with your current PRs. The thing I'm most worried about is what is the time from getting that PR merged - to seeing it in some conda channel?

@ccordoba12 I've got quite a lot of experience in getting VS 2008/2010 to work with a few popular Python extensions, so feel free to ping me on any very popular packages that are acting up and I'll see what I can do.

(@jakirkham I've moved my reply to your comment down here so that the action points can be read first)
Regarding the Windows builds - I am very well versed in supporting Windows Python extensions. In fact, I usually go so far as to stop using packages if I cannot get them to build for Windows. I am also well aware of the difficulty of providing binary packages on each of the different platforms (each having their own subtle pitfalls). And in fact, circling back to my quoted re-paraphrased question above, I totally agree that it is a nightmare trying to get different platforms to build *when you don't have any automated build system to verify package building.* Please see https://ci.appveyor.com/ for an excellent, free for open source, Windows automated build system that is very similar to Travis. I also have many scripts that help ease the burden of compiling within appveyor using conda.

I've been writing conda packages for a few years now and what originally steered me away from this repository was how it looked like what homebrew was doing - but wasn't. The lack of automated builds (apart from a proprietary system that is occasionally synced back to here), is a real bummer for trying to help curate this repository. As long as every PR requires some good natured people to go away and verify it on their machines it is going to be slow going in here.

patricksnape:
Also @jakirkham - great work on OSX builds! Top tip is that the first thing I always do when porting a new package is head over to homebrew and pretty much copy their script into Python - they always have great, working OSX scripts for most packages!

almarklein:
I changed the build process on Windows to use the method in @patricksnape 's repo. However, I cannot get it to work on my Windows VM's: on WinXP the `patch` fails because I did not have it, and the one I installed crashes. on Vista 64bit I get "the specified configuration type is missing" (it then carries on with the build, but fails to produce the libs). @patricksnape any ideas?

patricksnape:
Ah! I have had people complain about the lack of patch before, which is weird because it ships with conda - let me have a look and maybe I need to have a complete path to patch (I imagine it's because I'm not using the default command prompt). Vista 64-bit I'm not sure about, could you possible post the output in a gist and link it here?

Also - what version of Python?

msarahan:
Patch has changed a bit in the recent past.  It used to be a Python utility: https://code.google.com/p/python-patch/wiki/README

We ran into problems with Python 3 compatibility.  We changed Patch to be Gnuwin32's patch: http://gnuwin32.sourceforge.net/packages/patch.htm

This was not without issue: Windows thinks patch.exe requires elevation to an admin.  We fix that by including a .manifest file with it that tells Windows to not require elevation.  This is available as the package patch-2.5.9-1.  patch-2.5.9-0 does not include the manifest.

GnuWin32 patch seems to barf on line ending differences between Windows and Unix.  I have had the best luck simply forcing Unix line endings before patching, and always downloading source in Unix format (prefer tarballs to zip files.)  See https://github.com/conda/conda-build/blob/master/conda_build/source.py#L251  Note that this currently leaves .patch_unix files lying around.  This has been addressed in https://github.com/conda/conda-build/blob/msarahan/path_prepend/conda_build/source.py#L254 but not merged yet.

Updating conda-build and installing a current patch package might fix your problems.

msarahan:
Also, note that we're fixing the problem with different MSVC runtimes.  Each library like freeimage will require a section like:

```
build:
  features:
    - vc9     [win and py27]
    - vc10    [win and py34]
    - vc14    [win and py35]
```

Python itself will track those features, and install the library that is linked against the correct runtime.  I don't have updated recipes to post here yet, but will come back after the Anaconda 2.4 release tests this new system in the fire, so to speak.  The downside to this is that you (we) need to compile more libraries.  Pure Python packages alone are OK, but packages that depend on compiled libraries will each need separate builds.

You may see these showing up in the Anaconda repository.  Watch for the vc?? feature in your package lists.  Pretty soon, you'll want to start paying to attention to packages that do NOT include vc?? - these will be packages that need updating and rebuilding.

patricksnape:
@msarahan Thanks a lot for the patch information - I actually end up using patch quite a lot for conditional scripts, so this is incredibly useful for many of my repos.

Also, that is absolutely fantastic news about the vc* features. Genius way of trying to mitigate these issues within the current features framework. I will also endeavour to update all of my recipes to support this as soon as possible. First I need to get 3.5 builds up and running on Appveyor!

@msarahan Slightly off-topic - is there a Numpy 1.10 build for Python 3.5 on Windows yet? I checked a couple of days ago and there wasn't yet. Any idea when that is scheduled for?

almarklein:
@msarahan Great news. Can we already use the vc? stuff now, or will that come available with Anaconda 2.4?

msarahan:
vc? stuff is available now.  Make sure that your Conda version is updated.  I think 3.18.2 should be adequate.  Not all packages are built, though.  I have been working primarily on the Numpy 1.10 / Python 3.5 space.  These should all become available over the next day or two, and I'd very much appreciate help testing these builds.

Note that several binaries and libraries are moving into the Library folder (bin or lib or include).  This looks like a /usr linux folder, and makes installation simpler (just copy and keep native linux folder layout.)  If you have Python stuff on your PATH, you'll need to add the Library\bin folder manually, or else a lot of stuff might be broken.  Newer installers will do this for you.  Activating conda environments should also do this for you, as of conda-env 2.4.3, released Sunday.

For continuum-internal people, see recipes in https://github.com/ContinuumIO/anaconda (packages folder).  These will be rolled out to Conda-Recipes after more extensive testing.

For example, see the libtiff recipe for a library that defines features.  See the freetype package for a library that consumes other libraries (important detail: even when you don't use Python in a library, you must list it as a build time dependency, so that that the features will be resolved correctly on dependency libraries.)

almarklein:
Tried with vc?, but get same error: https://gist.github.com/almarklein/12aec846e3193f58e4a6
Maybe its me; I'm rather a noob when it comes to compiling stuff on Windows.

jakirkham:
Have you seen this ( http://stackoverflow.com/questions/2817869/error-unable-to-find-vcvarsall-bat )?

msarahan:
Almar, I hate to see you waste your time trying to figure out Windows build issues.  Please leave this with me, with the understanding that I need to prioritize Anaconda.  After that I'll be happy to fix this recipe for you.  If it is holding up your mission-critical work, let me know and I'll factor that in.

jakirkham:
Very late response to @patricksnape. I have carefully read your response ( https://github.com/conda/conda-recipes/pull/362#issuecomment-138995307 ) and have tried to distill a brief response here.

> Regarding the Windows builds - I am very well versed in supporting Windows Python extensions....

Which is why we are trying to solicit your help.

> I have a keen interested in imageio....

Great, it seems you already are being helpful.

> ...I totally agree that it is a nightmare trying to get different platforms to build when you don't have any automated build system to verify package building...

Agreed. CI is valuable. This undoubtedly will take a bit of work to set up with this repo, but it would be worthwhile. I use CI with most of my code and am aware of the options.

The main concern I worry about with CIs is how similar they (VMs) are to a real system in the wild. This concern only becomes bigger for me when I am thinking about something that is highly dependent on system configuration like a build framework. For instance, Travis uses Ubuntu. I have found Ubuntu and Enterprise Linux don't always play nicely together in often surprising ways. I worry in this particular case that we would become to dependent on CI and that it could give us a false sense of confidence. That doesn't mean we shouldn't pursue CI, but I don't think we should depend only on it.

> ...great work on OSX builds! Top tip is that the first thing I always do when porting a new package is head over to homebrew...

I am aware of the good work done over at homebrew. In this instance, these patches largely came from the FreeImage community itself.

almarklein:
@msarahan Oh that would be awesome. I just want to get a conda package for freeimage that works well on all platforms, so we can bring imageio to a higher level (and a new release, finally). If you can make that work, you are my hero (as well as anyone else here who has helped make this work)!

ukoethe:
@msarahan:
The introduction of the `vc?` features is great news. However, when I experimented with this idea, I found that the dependecy resolution algorithm occasionally fails to select the right version, see https://github.com/conda/conda/issues/1589. Has this problem been resolved in the meantime, or is there another reason it cannot occur in the context of `vc?` features?

> important detail: even when you don't use Python in a library, you must list it as a build time dependency, so that that the features will be resolved correctly on dependency libraries.

In my own recipe collection (see https://github.com/ilastik/ilastik-build-conda/tree/toolset-config), I introduced a recipe `visual-studio` which tracks the `vc?` feature and installs the appropriate runtime libraries. This avoids the need to always depend on Python.

>  Each library like freeimage will require a section like:
```
build:
  features:
    - vc9     [win and py27]
    - vc10    [win and py34]
    - vc14    [win and py35]
```

Repeating this in every recipe is error prone, especially when corresponding features need to be introduced for gcc4 vs. gcc5 as well. I propose to use jinja in order to define a variable that would allow a shorthand notation like:
```
build:
  features:
    - {{ toolset.features }}  # expands to 'vc9  [win and py27]' etc. as appropriate
````
See PR https://github.com/conda/conda-build/pull/578 for details.

msarahan:
@ukoethe 

> Has this problem been resolved in the meantime, or is there another reason it cannot occur in the context of vc? features?

I encounter this intermittently.  Usually the problem is that the appropriate package upstream doesn't exist, so it falls back to any package without the feature.  I think the long-term fix for that is to have no packages without features.  To be honest, I don't know how to best make this cut.  If you have suggestions, I'm wide open to them.

> In my own recipe collection (see https://github.com/ilastik/ilastik-build-conda/tree/toolset-config), I introduced a recipe visual-studio which tracks the vc? feature and installs the appropriate runtime libraries. This avoids the need to always depend on Python.

This is a cleaner solution.  Thanks.  I'll adapt our recipes to that.

> Repeating this in every recipe is error prone, especially when corresponding features need to be introduced for gcc4 vs. gcc5 as well. I propose to use jinja in order to define a variable that would allow a shorthand notation like:

This is much, much nicer!  Thanks.  I will get conda/conda-build#578 merged quickly and move to this approach.

ukoethe:
>  Usually the problem is that the appropriate package upstream doesn't exist, so it falls back to any package without the feature.

Another occasion where I saw the problem is when the newest version of a package (which didn't have the feature) was preferred over a slightly older version with the feature.

>  I think the long-term fix for that is to have no packages without features.

This would work, but is a tough guarantee to make, because you have no control over the set of channels which a particular install command will use.

>  If you have suggestions, I'm wide open to them.

See https://github.com/conda/conda/issues/1589#issuecomment-150154106 for a workaround until dependency resolution in the presence of features is fixed.

>  I'll adapt our recipes to that.

Out of curiosity: are the official Anaconda recipes published somewhere? I haven't been able to find them (e.g. `recipe` is missing in the packages' `info` directory).

patricksnape:
@msarahan @ukoethe Sorry for the late response, deadline is finally over so I'm just catching up. I'm a bit overwhelmed by the new VC14/VC10/VC9 feature system that is being used and I was wondering if there was any documentation I could read to get my head around what is the current best practise? Particularly in regards to @ukoethe's PR that seems to suggest moving towards a `jinja` template instead (which I fully support). I'd love to update all my recipes to have proper VC support under the new system (and in the mean time I could also look at the freeimage recipe) but I'm not sure where to start. Seems a lot of the info is spread out amongst conversations on a number of different repos - could we possible consolidate it somewhere, on the `wiki` or even just a `gist`? Or even here and I will write it up once I understand it!

almarklein:
@msarahan Any progress on this?

msarahan:
@almarklein thanks for the reminder. I have two things in front of this, but hope to work on this today.  @ukoethe and @patricksnape  I owe you answers as well, sorry. Coming soon.

msarahan:
I have submitted a PR against your branch, @almarklein - let me know if it does not show up.

@patricksnape I have started a draft of feature docs here: https://github.com/conda/conda/wiki/VC-features  Please ask if anything is not clear.  In particular, the template feature that @ukoethe has developed is not well documented, because I don't fully understand it (I don't understand where his template content gets defined).  @ukoethe if you could expand that section, I'd be thankful.

@ukoethe Official Anaconda recipes are internal and private.  Many of them are in a format that predates Conda.  Where we have already converted them to Conda recipes, we have permission to share them.  We can't share the unconverted recipes because we want to limit confusion about the format.  If you'd like to see the contents of a recipe, or have the recipe in conda-recipes updated, please ask.  We'll do what we can with the time that we have.

almarklein:
@msarahan is it supposed to build on a bare windows machine? I tried, but I get https://gist.github.com/almarklein/c14b34627e9219650a3c

On Linux I get a bunch of "error: 'memset' was not declared in this scope". Google says this has been a problem some years ago, but it ought to be fixed.

groutr:
@almarklein, do you have the Visual Studio 2010 compilers installed?

msarahan:
Yes, as Ryan say, it appears that you do not have Visual Studio 2010
installed.  Those executables are part of Visual Studio, and Python 3.4 is
restricted to building with VS 2010.

I haven't touched the linux part.  Were those errors there when you last
tried the recipe?

almarklein:
Aha; I thought the compilers would not be necessary because they are shipped via conda. The linux part may be broken because I tried on a different VM than before.

msarahan:
Runtimes are shipped, but not compilers.  It would be nice to ship compilers, but licensing makes that impossible right now.

For Linux, I recommend using centos 5 with gcc 4.4.  That's our standard Linux builder setup right now.  I am working on a docker container if you'd like.

jakirkham:
Linux worked before, right? I don't think the patch here should have broken it.

jakirkham:
@msarahan, is this going in [docker-images]( https://github.com/ContinuumIO/docker-images )

msarahan:
Once I clear it with the rest of the team, sure.  It looks to me like it should just be used in https://github.com/ContinuumIO/docker-images/blob/master/scripts/conda_build_in_docker.sh instead of the anaconda container.  Or one could use it directly instead...

almarklein:
So, will this lib be available in the main channels soon?

msarahan:
No, not by default. Conda-recipes does not directly feed the main channel.  If you want/need this package to be in defaults, please ask Crystal or Ilan.  Being in main implies a maintenance burden, and we can't assume that burden for all packages.

ccordoba12:
@msarahan, I think Freeimage is a dependency of imageio, which in turn is going to be a dependency of scikit-image.

So this needs to go to our repos :-)

msarahan:
Fair enough.  I'm sold.  I'll take this one on.

jakirkham:
I mean this almost done, no? Getting Windows support was the biggest missing piece. I guess we probably should recheck each platform.

msarahan:
Yes.  I'll recheck, merge, and then merge into the internal repo, and build out.

jakirkham:
Thanks @msarahan.

msarahan:
All platforms built fine for me: windows, linux x64 and OS X 10.7.

@almarklein one last bit of cleanup: I'd like to remove the VS2010 folder with project files.  We don't use them.  PR submitted against your branch.

ccordoba12:
@msarahan, thanks for taking care of this one. I promised @almarklein I'd do it but never found the time to do so ;-)

almarklein:
Yeah, thanks @msarahan, this is awesome. Being able to just do `conda install freeimage` is a game-changer for imageio.

jakirkham:
This is exciting!

jakirkham:
What other dependencies does imageio have, @almarklein?

almarklein:
@jakirkham numpy, that's it. We have some optional dependencies for more exotic formats (e.g. astropy, simpleitk), but I would not worry about these for now.

ukoethe:
@msarahan @jakirkham @stuarteberg @patricksnape @kalefranz
>  @ukoethe's PR seems to suggest moving towards a jinja template. 
> The template feature that @ukoethe has developed is not well documented, because I don't fully understand it (I don't understand where his template content gets defined).

I created issue https://github.com/conda/conda/issues/1959 to explain my suggestions in detail. Please comment!

PR #369, opened 2015-08-05
added pywavelets 0.3.0

mpesavento:


pywavelets/meta.yaml
```diff
@@ -0,0 +1,26 @@
+package:
+  name: pywavelets
+  version: "0.3.0"
+
+source:
+  git_url: https://github.com/rgommers/pywt.git
```
rgommers: This should now be https://github.com/PyWavelets/pywt.git

pywavelets/meta.yaml
```diff
@@ -0,0 +1,26 @@
...
+  run:
+    - python
+    - numpy
+
+test:
+  requires:
+    - nose
+
+about:
+  home: http://www.pybytes.com/pywavelets/
```
rgommers: We've migrated away from ``pybytes.com``. I suggest using https://github.com/PyWavelets/pywt here. Or otherwise http://pywavelets.readthedocs.org is also OK.

rgommers:
Thanks @mpesavento, looks good to me apart from the URLs I commented on. (disclaimer: I'm not a conda recipe expert)

mpesavento:
Updated the URLs after testing the build with the new repo link. Looks good from my side.

PR #372, opened 2015-08-08
Add GCC 5 package

frol:
This package is based on refined scripts for GCC 4.8 (#279) and tested on: CentOS 5.11, CentOS 6.6, Ubuntu 12.04, Ubuntu 14.04.

This pull request also contains libgcc-5 package similar to libgcc package.

NOTE: GCC 5 introduced new versioning, so they use the second number in the version for bugfix-releases in contrast to 4.x version, where  bugfix-releases were released with the third number.

WARNING: GCC 5 has introduced new ABI and my package builds GCC 5.2 with new ABI enabled by default. Thus, those who link to other libraries from Anaconda repo will have to define `_GLIBCXX_USE_CXX11_ABI=0` macro via, for example, `CXXFLAGS`. For cmake-driven projects, it can be passed with the following command:

```
$ cmake -D CMAKE_C_FLAGS=-D_GLIBCXX_USE_CXX11_ABI=0 -D CMAKE_CXX_FLAGS=-D_GLIBCXX_USE_CXX11_ABI=0 ..
```

In our projects we moved to new ABI since our dependencies list is not that long. We only rebuilt Boost, though we had to do this for Python3 support anyway. I will make another pull-request for Boost package shortly.

### DIFF
--- a/gcc-5/build.sh
+++ b/gcc-5/build.sh
@@ -3,6 +3,7 @@
 ln -s "$PREFIX/lib" "$PREFIX/lib64"
 
 if [ "$(uname)" == "Darwin" ]; then
+    export DYLD_LIBRARY_PATH="$PREFIX/lib/"
     export LDFLAGS="-Wl,-headerpad_max_install_names"
     export BOOT_LDFLAGS="-Wl,-headerpad_max_install_names"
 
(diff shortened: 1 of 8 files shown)


frol:
You can try it:

```
$ conda install -c salford_systems gcc-5
```

frol:
Please, review, but don't merge it yet. I've found `make install-strip` and testing it now, so probably the hack with `find-strip` will be removed.

frol:
`make install-strip` makes this package even more compact - 266MB unpacked and 78MB tar.bz2. I've pushed the patch.

frol:
I've tested this package on MacOS X and Linux x64 platforms and it builds and works fine.

PR #373, opened 2015-08-08
Improved GCC 4.8 package

frol:
Added a step to strip debug symbols on Linux that makes the package 60% smaller (860MB -> 290MB unpacked and 280MB -> 78MB tar.bz2).

libgcc package had unnecessary run dependencies.

This pull request was inspired by #279

### DIFF
--- a/libgcc/meta.yaml
+++ b/libgcc/meta.yaml
@@ -2,6 +2,9 @@ package:
   name: libgcc
   version: 4.8.5
 
+build:
+  number: 3
+
 requirements:
   build:
     - gcc 4.8.5
--- a/libgcc/meta.yaml
+++ b/libgcc/meta.yaml
@@ -5,13 +5,6 @@ package:
 requirements:
   build:
     - gcc 4.8.5
-  run:
-    # I'm unsure if these are all still needed but it doesn't hurt to have them.
-    - gmp >=4.2
-    - mpfr >=2.4.0
-    - mpc >=0.8.0
-    - isl
-    - cloog 0.18.0
 
 build:
   always_include_files:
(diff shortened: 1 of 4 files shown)


frol:
@stuarteberg @asmeurer It would be nice if you could review my changes.

frol:
Please, don't merge it yet. I've found `make install-strip` and testing it now, so probably the hack with `find-strip` will be removed.

frol:
I've tested `make install-strip` on `gcc-5` package and `gcc-4.8` is rebuilding now, but it seems it should work and make the package even more compact than my find-strip hack.

frol:
@stuarteberg @asmeurer My pull request is nice and clean now. The package is 65MB tar.bz2 instead of 250MB!

ccordoba12:
I like this improvement. If people needs to build something with debugging info, they can use system `gcc`.

Conda is package manager after all, not a development tool.

openvest:
This is a great. Thanks, I'd been treating the gcc package as a necessary
evil.  Not it's wart free and I feel better putting it in front of the
client again.

--
Phil

frol:
@ccordoba12 @asmeurer Do you see any showstoppers for this pull request?

ccordoba12:
I'm sorry but you'll have to wait for @asmeurer answer. He's the only one who has worked on this package.

frol:
@asmeurer Please, review this PR.

asmeurer:
Can you bump the build numbers?

frol:
@asmeurer Done.

libgcc/meta.yaml
```diff
@@ -2,16 +2,12 @@ package:
   name: libgcc
   version: 4.8.5
 
+build:
+  number: 3
```
asmeurer: This was wrong (there is already a build section).

frol: Ouch, I'm sorry, I missed that.

Issue #341, opened 2015-07-06
How about splitting matplotlib into matplotlib and matplotlib-qt?


frol:
I don't really like the idea of getting `pyqt` and `qt` when I clearly don't need them. It seems to me that simple meta-packaging will right the ship.

ccordoba12:
You can just do

    conda install --no-deps matplotlib

to not grab `pyqt` as part of mpl for now.

frol:
I have a package that optionally depends on matplotlib (I set it up as a meta-package with `-matplotlib` suffix) and it makes installation instructions weird.

By the way, I tried to remove pyqt and it turned out that it is not enough!
```
ImportError: No module named PyQt4
```

I had to edit matplotlib package's `mpl-data/matplotlibrc` (I commented out `backend: Qt4Agg`) to make it work.

The `pyqt` dependency is not only annoying 40MB of traffic and 180MB on disk, but also it requires libXext.so.6 library on Linux, which is a part of X11, which is not what I want on my server and docker container.

frol:
Is there a specific reason not to move Qt dependency to separate meta-package? In fact, it seems that the default matplot backend is `agg` and not `Qt4Agg`. Why was it decided to use Qt, which seems to cause quite some troubles, as a default in the recipe?

frol:
cc @ccordoba12 @asmeurer

frol:
At least I figured out how to switch matplotlib backend without patching matplotlib package:

> matplotlib uses `matplotlibrc` configuration files to customize all kinds of properties, which we call **rc settings** or **rc parameters**. You can control the defaults of almost every property in matplotlib: figure size and dpi, line width, color and style, axes, axis and grid properties, text and font properties and so on. matplotlib looks for `matplotlibrc` in three locations, in the following order:
> 
> 1. **`matplotlibrc` in the current working directory, usually used for specific customizations that you do not want to apply elsewhere.**
> 2. It next looks in a user-specific place, depending on your platform:
> 
>   On Linux, it looks in `.config/matplotlib/matplotlibrc` (or `$XDG_CONFIG_HOME/matplotlib/matplotlibrc`) if you’ve customized your environment.
> 
>   On other platforms, it looks in `.matplotlib/matplotlibrc`.
> See .matplotlib directory location.
[...]

PR #374, opened 2015-08-08
Add Matplotlib no-GUI package

frol:
no Tk, no Qt, no GTK, only image-generation backends are left for iPython Notebooks.

There is a discussion on #341

### DIFF
--- a/matplotlib-nogui/build.sh
+++ b/matplotlib-nogui/build.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+if [ `uname` == Linux ]; then
+    pushd $PREFIX/lib
+    ln -s libtcl8.5.so libtcl.so
+    ln -s libtk8.5.so libtk.so
+    popd
+fi
+
+if [ `uname` == Darwin ]; then
+    sed s:'#ifdef WITH_NEXT_FRAMEWORK':'#if 1':g -i src/_macosx.m
+fi
+
+cp setup.cfg.template setup.cfg || exit 1
+
+sed s:/usr/local:$PREFIX:g -i setupext.py
+
+$PYTHON setup.py install
+
+rm -rf $SP_DIR/PySide
+rm -rf $SP_DIR/__pycache__
+rm -rf $PREFIX/bin/nose*
+
--- a/matplotlib-nogui/build.sh
+++ b/matplotlib-nogui/build.sh
@@ -1,23 +1,19 @@
 #!/bin/bash
 
-if [ `uname` == Linux ]; then
-    pushd $PREFIX/lib
-    ln -s libtcl8.5.so libtcl.so
-    ln -s libtk8.5.so libtk.so
-    popd
-fi
-
-if [ `uname` == Darwin ]; then
-    sed s:'#ifdef WITH_NEXT_FRAMEWORK':'#if 1':g -i src/_macosx.m
+if [[ "$(uname)" == "Darwin" ]]
+then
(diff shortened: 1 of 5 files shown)


ccordoba12:
@tacaswell, could you take a look at this recipe?

Is this something you think it'd be useful to have here?

frol:
You can test it from our repo:

```
$ conda install -c salford_systems matplotlib-nogui
```

We have it for linux-64 and osx.

matplotlib-nogui/disable_plugins.patch
```diff
@@ -0,0 +1,39 @@
+--- setupext.py
++++ setupext.py
+@@ -1050,22 +1050,8 @@
+     name = "qhull"
+ 
+     def check(self):
+-        self.__class__.found_external = True
```
tacaswell: Why do you need to remove the qhull checking?

matplotlib-nogui/build.sh
```diff
@@ -0,0 +1,19 @@
...
+fi
+
+cp setup.cfg.template setup.cfg || exit 1
+
+"${SED_INPLACE[@]}" "s:/usr/local:$PREFIX:g" setupext.py
+
+"$PYTHON" setup.py install
+
+rm -rf "$SP_DIR/PySide"
+rm -rf "$SP_DIR/__pycache__"
```
tacaswell: Isn't conda clever enough to not pick up these files on it's own?

tacaswell:
There seems to be a bunch of un-related changes.

Are there version of python that don't have tk?  I thought it was part of the standard library.

matplotlib-nogui/disable_plugins.patch
frol: There are several reasons:

1. there is no conda package for qhull (or I couldn't find it)
1. since matplotlib has a local copy and it seems to work fine, we had better stick to it explicitly

P.S. Unfortunately, I don't recall at the top of my head what was the exact issue.

matplotlib-nogui/build.sh
frol: These were copied from the original matplotlib package in conda-recipes.

frol:
@tacaswell As I recall, when I build matplotlib with Tk, it requires libX11.so, which is not what I wanted on our headless servers, where we run IPython Notebooks.

janssen:
Yes, please.  I'm trying to use matplotlib with Kivy, and it's a real pain to have to delete all the Qt dependencies after-the-fact.  There should be separate matplotlib-with-qt, and the core matplotlib should have no dependencies on qt.

matplotlib-nogui/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - numpy
+    - python-dateutil
+    - freetype            [unix]
+    - pytz
+    - pyparsing
+    - py2cairo            [linux and py2k]
+    - libpng              [unix]
+
+about:
+  home: http://matplotlib.sourceforge.net/
```
tacaswell: This is no longer correct.

tacaswell:
The main mpl recipe should also be changed to depend on this version and this should be updated to reflect the mpl / mpl-tests split.  I think @mdboom and I should take over dealing with this?

mcg1969:
@tacaswell, yes, absolutely. It would be great if you could take over the conda recipes. I've already reached out to @mdboom about handling conda package building. Feel free to reach out to me for more information.

PR #375, opened 2015-08-08
Add Python 3 support for Boost package

frol:
- [x] Added support of Python 3 (It has to be built in Python 2 and Python 3 environments now.)
- [x] Fixed some configuration options for OS X and Linux.
- [x] Upgraded to 1.58.0
- [x] Tested on CentOS 6.6, Ubuntu 14.04, OS X
- [x] Add `icu` support for OS X

### DIFF
--- a/boost/build.sh
+++ b/boost/build.sh
@@ -22,6 +22,8 @@ if [ "$(uname)" == "Darwin" ]; then
 
     ./bootstrap.sh \
         --prefix="${PREFIX}" \
+        --with-python="${PYTHON}" \
+        --with-icu="${PREFIX}" \
         | tee bootstrap.log 2>&1
 
     ./b2 -q \
--- a/boost/build.sh
+++ b/boost/build.sh
@@ -8,36 +8,54 @@
 # Hints for OSX:
 # http://stackoverflow.com/questions/20108407/how-do-i-compile-boost-for-os-x-64b-platforms-with-stdlibc
 
-mkdir -vp ${PREFIX}/bin;
+set -x -e
 
-if [ `uname` == Darwin ]; then
+if [ "$(uname)" == "Darwin" ]; then
     MACOSX_VERSION_MIN=10.8
     CXXFLAGS="-mmacosx-version-min=${MACOSX_VERSION_MIN}"
     CXXFLAGS="${CXXFLAGS} -std=c++11 -stdlib=libc++"
     LINKFLAGS="-mmacosx-version-min=${MACOSX_VERSION_MIN} "
-    LINKFLAGS="${LINKFLAGS} -stdlib=libc++ -L${LIBRARY_PATH}"
+    LINKFLAGS="${LINKFLAGS} -stdlib=libc++"
 
     ./bootstrap.sh \
-      --prefix="${PREFIX}/" --libdir="${PREFIX}/lib/" \
-      | tee bootstrap.log 2>&1
+        --prefix="${PREFIX}" \
+        | tee bootstrap.log 2>&1
 
     ./b2 \
-      variant=release address-model=64 architecture=x86 \
-      threading=multi link=shared toolset=clang include=${INCLUDE_PATH} \
-      cxxflags="${CXXFLAGS}" linkflags="${LINKFLAGS}" \
-      -j$(sysctl -n hw.ncpu) \
-      install | tee b2.log 2>&1
+        variant=release \
+        address-model=64 \
+        architecture=x86 \
+        threading=multi \
+        link=shared \
+        toolset=clang \
+        cxxflags="${CXXFLAGS}" \
+        linkflags="${LINKFLAGS}" \
+        -j"$(sysctl -n hw.ncpu)" \
+        install | tee b2.log 2>&1
 fi
 
-if [ `uname` == Linux ]; then
-  ./bootstrap.sh \
-    --prefix="${PREFIX}/" --libdir="${PREFIX}/lib/" \
-    | tee bootstrap.log 2>&1
+if [ "$(uname)" == "Linux" ]; then
+    ./bootstrap.sh \
+        --prefix="${PREFIX}" \
+        --with-python="${PYTHON}" \
+        --with-python-root="${PREFIX} : ${PREFIX}/include/python${PY_VER}m ${PREFIX}/include/python${PY_VER}" \
+        --with-icu="${PREFIX}" \
+        | tee bootstrap.log 2>&1
 
-  ./b2 \
-    variant=release address-model=${ARCH} architecture=x86 \
-    threading=multi link=shared toolset=gcc include=${INCLUDE_PATH} \
-    -j${CPU_COUNT} \
-    install | tee b2.log 2>&1
+    ./b2 -q -d0 \
+        variant=release \
+        address-model="${ARCH}" \
+        architecture=x86 \
+        debug-symbols=off \
+        threading=multi \
+        runtime-link=shared \
+        link=shared,static \
+        toolset=gcc \
+        python="${PY_VER}" \
+        cflags="-O3 -I${PREFIX}/include" \
+        cxxflags="-O3 -I${PREFIX}/include" \
+        linkflags="-L${PREFIX}/lib" \
+        --layout=system \
+        -j"${CPU_COUNT}" \
+        install | tee b2.log 2>&1
 fi
-
--- a/boost/build.sh
+++ b/boost/build.sh
@@ -10,24 +10,29 @@
 
 set -x -e
 
+INCLUDE_PATH="${PREFIX}/include"
+LIBRARY_PATH="${PREFIX}/lib"
+
 if [ "$(uname)" == "Darwin" ]; then
     MACOSX_VERSION_MIN=10.8
     CXXFLAGS="-mmacosx-version-min=${MACOSX_VERSION_MIN}"
     CXXFLAGS="${CXXFLAGS} -std=c++11 -stdlib=libc++"
-    LINKFLAGS="-mmacosx-version-min=${MACOSX_VERSION_MIN} "
-    LINKFLAGS="${LINKFLAGS} -stdlib=libc++"
+    LINKFLAGS="-mmacosx-version-min=${MACOSX_VERSION_MIN}"
+    LINKFLAGS="${LINKFLAGS} -stdlib=libc++ -L${LIBRARY_PATH}"
 
     ./bootstrap.sh \
         --prefix="${PREFIX}" \
         | tee bootstrap.log 2>&1
 
-    ./b2 \
+    ./b2 -q \
         variant=release \
         address-model=64 \
         architecture=x86 \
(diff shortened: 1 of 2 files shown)


boost/meta.yaml
```diff
@@ -17,6 +17,8 @@ requirements:
     - libpython         [win]
```
ccordoba12: Why `python` is not a build dependency?

boost/build.sh
```diff
@@ -8,36 +8,54 @@
...
+        | tee bootstrap.log 2>&1
+
+    ./b2 -q -d0 \
+        variant=release \
+        address-model="${ARCH}" \
+        architecture=x86 \
+        debug-symbols=off \
+        threading=multi \
+        runtime-link=shared \
+        link=shared,static \
```
ccordoba12: Why do we need to create static libs too?

boost/meta.yaml
frol: What do you mean? The first line in build dependencies is `python`. Nevertheless, I haven't touched Windows-specific configs at all. I tested everything on CentOS 6.6, Ubuntu 14.04, OS X, though.

boost/build.sh
```diff
@@ -8,36 +8,54 @@
...
+        toolset=gcc \
+        python="${PY_VER}" \
+        cflags="-O3 -I${PREFIX}/include" \
```
ccordoba12: Why not use the `include` switch instead of adding build includes to `cflags / cxxflags`?

It seems a bit ugly to me :-)

ccordoba12: Also, why do you want to optimize this with `03`. Such optimizations can cause problems.

boost/build.sh
```diff
@@ -8,36 +8,54 @@
...
+        cxxflags="-O3 -I${PREFIX}/include" \
+        linkflags="-L${PREFIX}/lib" \
+        --layout=system \
```
ccordoba12: What does this `layout` switch do?

boost/build.sh
frol: Why shouldn't we? Anyway, I can change it back to shared-only. I copied these options from Arch Linux packaging.

boost/meta.yaml
ccordoba12: Ah, sorry. I haven't noticed it. I think it wasn't there before, so I don't know why is not reflected on the diff :-)

boost/build.sh
ccordoba12: I say it because it makes the package bigger for no good reason.

boost/build.sh
```diff
@@ -8,36 +8,54 @@
...
-    variant=release address-model=${ARCH} architecture=x86 \
-    threading=multi link=shared toolset=gcc include=${INCLUDE_PATH} \
-    -j${CPU_COUNT} \
-    install | tee b2.log 2>&1
+if [ "$(uname)" == "Linux" ]; then
+    ./bootstrap.sh \
+        --prefix="${PREFIX}" \
+        --with-python="${PYTHON}" \
+        --with-python-root="${PREFIX} : ${PREFIX}/include/python${PY_VER}m ${PREFIX}/include/python${PY_VER}" \
+        --with-icu="${PREFIX}" \
```
ccordoba12: Have you checked that `icu` is detected correctly?

boost/build.sh
frol: `--include` switch doesn't work like you think (I expected it would work like that also, BTW). It specifies where Boost will install (!) the libraries, not where it will search on build step. The same thing for `--libsdir`.

ccordoba12: Well, i used it before to make Boost detect the icu headers.

frol: `O3` is taken from Arch Linux package, which seems to work for a number of people. Let's optimize as much as possible unless we get a bug report here, this is my way of thinking.

ccordoba12: Problem is optimizations bugs are hard to detect and/or report. Well, `03` shouldn't cause much harm, but for things to work reliably everything should be compiled with `03` (at least that's how Gentoo works :-).

Since we're not doing that, I'd prefer to **not** add optimizations at all.

boost/build.sh
```diff
@@ -8,36 +8,54 @@
...
-mkdir -vp ${PREFIX}/bin;
+set -x -e
 
-if [ `uname` == Darwin ]; then
+if [ "$(uname)" == "Darwin" ]; then
     MACOSX_VERSION_MIN=10.8
     CXXFLAGS="-mmacosx-version-min=${MACOSX_VERSION_MIN}"
     CXXFLAGS="${CXXFLAGS} -std=c++11 -stdlib=libc++"
     LINKFLAGS="-mmacosx-version-min=${MACOSX_VERSION_MIN} "
-    LINKFLAGS="${LINKFLAGS} -stdlib=libc++ -L${LIBRARY_PATH}"
```
ccordoba12: Why did you remove this?

boost/build.sh
```diff
@@ -8,36 +8,54 @@
...
+    LINKFLAGS="${LINKFLAGS} -stdlib=libc++"
 
     ./bootstrap.sh \
-      --prefix="${PREFIX}/" --libdir="${PREFIX}/lib/" \
```
ccordoba12: And also removed this?

boost/build.sh
frol: It is actually `system` by default on Linux anyway, but I just stated it explicitly.

> system - Builds an install without the Boost standard names, and does not install includes to a versioned sub-directory. This is intended for system integrators to build for packaging of distributions.

You can read all options from boost sources or here is dated help that I could find online: http://www.boost.org/doc/libs/1_33_1/Jamfile

boost/build.sh
frol: Yes, it is. Once I specified `"$PREFIX"` explicitly, it gets detected correctly.

boost/build.sh
frol: Because conda-build doesn't provide any LIBRARY_PATH env variable anyway.

boost/build.sh
frol: `libdir` specifies where libraries will be installed, but if you set `prefix` it automatically set `libdir` to `${PREFIX}/lib` and `include` to `${PREFIX}/include`

boost/build.sh
frol: Should I remove `O3` completely, or downgrade to `O2` will be fine?

frol: @ccordoba12 I have no idea how `include` helped you with `icu` location. Here is the help:

```
Usage: ./bootstrap.sh [OPTION]...

Defaults for the options are specified in brackets.

Configuration:
  -h, --help                display this help and exit
  --with-bjam=BJAM          use existing Boost.Jam executable (bjam)
                            [automatically built]
  --with-toolset=TOOLSET    use specific Boost.Build toolset
                            [automatically detected]
  --show-libraries          show the set of libraries that require build
                            and installation steps (i.e., those libraries
                            that can be used with --with-libraries or
                            --without-libraries), then exit
  --with-libraries=list     build only a particular set of libraries,
                            describing using either a comma-separated list of
                            library names or "all"
                            [all]
  --without-libraries=list  build all libraries except the ones listed []
  --with-icu                enable Unicode/ICU support in Regex
                            [automatically detected]
  --without-icu             disable Unicode/ICU support in Regex
  --with-icu=DIR            specify the root of the ICU library installation
                            and enable Unicode/ICU support in Regex
                            [automatically detected]
  --with-python=PYTHON      specify the Python executable [python]
  --with-python-root=DIR    specify the root of the Python installation
                            [automatically detected]
  --with-python-version=X.Y specify the Python version as X.Y
                            [automatically detected]

Installation directories:
  --prefix=PREFIX           install Boost into the given PREFIX
                            [/usr/local]
  --exec-prefix=EPREFIX     install Boost binaries into the given EPREFIX
                            [PREFIX]

More precise control over installation directories:
  --libdir=DIR              install libraries here [EPREFIX/lib]
  --includedir=DIR          install headers here [PREFIX/include]
```

It states clearly, that it is used only for a deployment of resulting files.

boost/build.sh
ccordoba12: Ok, no problem on leaving it :-)

boost/build.sh
ccordoba12: I used `--include` not `--includedir`, as you can see in the diff :-)

ccordoba12: Also, please remove all optimizations, i.e. not `03` nor `02`

boost/build.sh
ccordoba12: Ok, thanks for the explanation

boost/build.sh
ccordoba12: Right, I always forget that!

boost/build.sh
ccordoba12: So, please remove `static` libs.

boost/build.sh
frol: `include` is indeed available, so I changed it back. I also removed `O3` and `static`.

boost/build.sh
frol: Done

boost/build.sh
```diff
@@ -8,36 +8,58 @@
...
+
+    ./b2 -q \
+        variant=release \
+        address-model=64 \
+        architecture=x86 \
+        debug-symbols=off \
+        threading=multi \
+        link=shared \
+        toolset=clang \
+        include="${INCLUDE_PATH}" \
```
ccordoba12: Remember that `INCLUDE_PATH` is not set anymore. This has to be `"${PREFIX}"/include`

ccordoba12: Sorry, I just saw you set it above.

ccordoba12:
@frol, thanks for the improvements. I'll test this PR this afternoon before merging.

boost/build.sh
frol: Question about optimizations bothered me so I created another issue to discuss it: #376

frol:
@ccordoba12 Have you had a chance to test it? Have you run into any issues?

ccordoba12:
Could you please add the same options for `icu` and `python` that you're using for Linux to Mac too?

I think that's the only thing left before merging.

frol:
@ccordoba12 Those changes were unnecessary for OS X, it seems that autodiscovery worked better there than on Linux. Nevertheless, I set those explicitly now. I also bumped Boost version to 1.59 (released on August 13th). I'm testing the new builds now:

```
$ conda install -c salford_systems boost=1.59
```

frol:
@ccordoba12 All my tests passed. Please, note that my focus is on Linux. I have no project on OS X yet.

ccordoba12:
I have. I tested your recipes and they're working fine :-)

frol:
@ccordoba12 How long does it usually take for a package to get to the official anaconda channel?

ccordoba12:
I'm charge in the Boost packages, but I'm very busy right now. I think I'll do the update in three weeks or a month.

PR #379, opened 2015-08-12
jq: Fix dead URL

jakirkham:


jakirkham:
@ccordoba12, turns out the URL no longer works. So, fixed to point to working source. Builds fine. Sample packages in my channel if you are curious.

PR #384, opened 2015-08-17
Adds recipe to build pylibtiff

jakirkham:
This works on both Mac and Linux. It was tested on Mac OS 10.9 and CentOS 6.3.

This is patched to simplify finding `libtiff`. This patch is being submitted as a PR ( https://github.com/pearu/pylibtiff/pull/36 ).

### DIFF
--- a/pylibtiff/build.sh
+++ b/pylibtiff/build.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+set -e
+
+python setup.py install --prefix=$PREFIX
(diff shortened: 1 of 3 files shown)


ilanschnell:
Great!  I'm happy you found a way to avoid setting `LD_LIBRARY_PATH`, which would have been a terrible hack, and would have caused all types of other problems for users.

jakirkham:
Me too. Thanks for your help. I just wish `ctypes` had better documentation about finding and loading libraries.

PR #386, opened 2015-08-18
Switch from using `gcc` to `libgcc` as a runtime requirement

jakirkham:
I tried to get everything using `gcc` as a runtime dependency to use `libgcc` instead. It turned out only one package was doing this.

### DIFF
--- a/tig/meta.yaml
+++ b/tig/meta.yaml
@@ -13,7 +13,7 @@ requirements:
     - gcc
   run:
     - ncurses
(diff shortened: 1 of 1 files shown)


jakirkham:
@asmeurer, do you have any thoughts on merging this?

ccordoba12:
It seems very simple. Thanks :-)

jakirkham:
Thanks @ccordoba12.

PR #390, opened 2015-08-21
Add OpenBLAS, NumPy, and SciPy

jakirkham:
This is based off of work done by @abergeron in this PR ( https://github.com/conda/conda-recipes/pull/331 ). This is also based off the OpenBLAS recipes here ( https://github.com/jakirkham/nanshe-build/ ), which build one the work of @hajs here ( https://github.com/hajs/pylodger ).

* Ensures LAPACK is built in OpenBLAS.
* Provides a standard API for linking the BLAS and LAPACK within OpenBLAS.
* Combines the recipes `openblas` and `openblas-lib`.
* Respects 32-bit and 64-bit when specified by conda.
* Speeds up the OpenBLAS build.
* Updates SciPy to 0.16.0. However, the old version is provided as the previous to last commit.
* Ensures that gfortran libraries are picked up correctly on the path.
* Makes `libgcc` a dependency so all needed libraries will be included.
* Avoids unfriendly linking arguments that hurt the build of NumPy and SciPy.
* OpenBLAS is fixed at 1 thread. However, this can be change by setting the environment variable OPENBLAS_NUM_THREADS.

This has been tested on Mac ( 10.9 and 10.10 ) and Linux ( CentOS 6.3 ). Also, it has been tested on the following architectures: Sandybridge, Haswell, and Broadwell.

ping @ilanschnell

Closes #331.


abergeron:
Nice!

Sorry for the lack of feedback on the other PR I was away for while.

As for merging the metapackage and the "real" one, that was my first attempt, but it didn't work.  I think that this was before I figure that I needed to add a tag to the build string for numpy and scipy.

Anyway if it works like this, it's definitely better.

ilanschnell:
Great.  The default conda repository repo.continuum.io now also has openblas on Linux, and numpy and scipy linked to it by default (instead of ATLAS).

jakirkham:
@abergeron, yeah, sorry to open this other one. I did include your original work though. If you have any feedback, feel free to share.

@ilanschnell, cool, it was looking like you guys were moving in that direction. I was meaning to get these changes out sooner to avoid duplication of the effort. Sorry, that didn't happen completely. Hopefully, it is still of use.

Also, I added some more information about how I built on Mac and Linux at the top. For better or worse I have gained a fair bit of experience building this package, if you guys have difficulties getting it to work, please let me know and I will try to help.

jakirkham:
It would be nice to see OpenBLAS for Mac as Accelerate isn't fork safe. Using that with Python, which relies heavily on forking, ends up being a nightmare. Also, Apple, when last asked, has no interest in exploring a fix for this.

asmeurer:
Why don't you use the `gcc` conda package?

jakirkham:
@asmeurer, fair question. I tried to and it crashed during the build. I don't remember the exact error off the top of my head, but it seemed like a linking issue.

jakirkham:
@asmeurer, here are the build logs on Mac OS 10.9 where the error arose ( https://gist.github.com/jakirkham/830a4b48165f13de26bb ). In the first case, `gcc` was added as a dependency. You may notice that it still uses `clang`. In the second case, I have explicitly told it to use the `gcc` toolset from `conda`; in particular, the one built from the recent PR. However, it shouldn't matter as there were no real changes to the Mac build. In both cases, the build errors.

jakirkham:
@ilanschnell, on Linux I am unable to download `libgfortran`. It appears to be in the channel, but when I click on the link for either of the packages I get a 404 error, which is the same error `conda` gives. Just thought you should know.

ilanschnell:
@jakirkham `libgfortran` is in the default channels on Linux.  For example on x86_64 the location is
http://repo.continuum.io/pkgs/free/linux-64/libgfortran-1.0-0.tar.bz2
I have no problems downloading the file (and I also verified the MD5).

jakirkham:
@ilanschnell, I added some screenshots above.

You are correct that installing works correctly. In fact, installing say `numpy` works correctly and grabs `openblas` and `libgfortran`.

However, in this particular build of `bokeh` it breaks. ( https://github.com/bokeh/bokeh/issues/2488#issuecomment-139939084 ) I find it a little peculiar that I see the same error in the browser, but it may be coincidental.

PR #397, opened 2015-08-27
colorlog: new package

frol:
It allows colors to be placed in the format logging string, which is mostly useful when paired with a StreamHandler that is outputting to a terminal. This is accomplished by added a set of terminal color codes to the record before it is used to format the string.

Details about the package are here: https://github.com/borntyping/python-colorlog


ilanschnell:
Thanks!

PR #398, opened 2015-08-27
invoke: new package

frol:
Invoke is a Python (2.6+ and 3.3+) task execution tool & library, drawing inspiration from various sources to arrive at a powerful & clean feature set.

It is a future part of Fabric 2.0. Though it still hasn't reached release state, it works fine enough for us and many other DevOps.

### DIFF
--- a/invoke/bld.bat
+++ b/invoke/bld.bat
@@ -7,7 +7,7 @@ if "%PY_VER%" == "2.7" rmdir /S /Q "invoke\vendor\yaml3"
 else rmdir /S /Q "invoke\vendor\yaml2"
 
 :: Workaround conda-build issue #251: https://github.com/conda/conda-build/issues/251
-ECHO "program_run = program.run">>invoke\main.py
+ECHO program_run = program.run >>"invoke\main.py"
 
 "%PYTHON%" setup.py install
 if errorlevel 1 exit 1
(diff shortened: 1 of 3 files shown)


msarahan:
I merged another invoke PR without seeing this one. Do you mind taking a look and seeing if any of your work here would improve the recipe I merged?

frol:
@msarahan My recipe has the following features in contrast to the one you have merged:

1. Mine is based on the git master branch which has a number of fixes, which are not released since September. The merged one uses even older 0.10.1 version (released in April 2015).
2. Mine builds for Python 2 and 3, while the merged one won't get built for Python 2 due to the fact that Invoke has yaml3 with Python 3 syntax (the details are provided in my `build.sh`, `bld.bat`)
3. Entrypoint workaround that I have applied in my recipe is necessary for the newer version of Invoke (the author has modified the entrypoint code so it became class-based instead of a function).

I have been using my recipe on Windows, Linux, OS X with Python 2.7 and 3.4+ everywhere since I have PRed it, so for 5 months already, and it works fine. However, you may not like to merge a git-based recipe.

msarahan:
Those are nice improvements that I'd like to merge. We can't merge simple git recipes without some kind of version pin though: it is important that a recipe doesn't change version without manually changing the recipe.  I see the advantages that a "rolling" recipe could offer, though.  Perhaps we could explore a conda build enhancement that allows specification of version to build on the command line, and corresponding template placeholders in recipes.  

Do you mind pinning to the 0.11.1 tag for now?

frol:
@msarahan 0.11.1 is not compatible with my recipe. A number of changes were made in the master, a number of bugs were fixed, which is why I used master branch in the first place. Since nobody really use it yet, we can just wait for 0.12 release (hopefuly it will be released at some point of time...)

PR #399, opened 2015-08-28
nuitka: new package

frol:
> Nuitka is a Python compiler.

> It's fully compatible with Python 2.6, 2.7, 3.2, 3.3, and 3.4.

> You feed it your Python app, it does a lot of clever things, and spits out an executable or extension module.

http://nuitka.net/pages/overview.html


msarahan:
Thanks!

Issue #302, opened 2015-05-17
Cannot build gcc or libgcc on MacOS X (can't find ISL?)


rwest:
I'm trying to make a conda build recipe for something that needs a fortran compiler, so wanted `gcc` as a build dependency and `libgcc` as a runtime dependency. I am currently on a Mac (but would eventually like a cross-platform build). I am new to conda, so apologies if I've done something stupid:

The first strange thing is that `build libgcc` claims to be building the libraries for 4.8.4, but in fact downloads and tries to build gcc 4.9.1 as a dependency:
```
conda build libgcc
Removing old build directory
Removing old work directory
BUILD START: libgcc-4.8.4-1
Fetching package metadata: ......
Missing dependency gcc, but found recipe directory, so building gcc first
Missing dependency gcc, but found recipe directory, so building gcc first
Missing dependency gcc, but found recipe directory, so building gcc first
Removing old build directory
Removing old work directory
BUILD START: gcc-4.9.1-0
Fetching package metadata: ......
Solving package specifications: .
The following NEW packages will be INSTALLED:

    cloog: 0.18.1-0
    gmp:   5.1.2-6 
    isl:   0.12.2-1
    mpc:   1.0.1-0 
    mpfr:  3.1.2-0 

Linking packages ...
[      COMPLETE      ]|#############################################################################################| 100%
Removing old work directory
Source cache directory is: /Users/rwest/anaconda/conda-bld/src_cache
Found source in cache: gcc-4.9.1.tar.bz2
Extracting download
etc.
```
It then fails anyway:
```
Package: gcc-4.9.1-0
source tree in: /Users/rwest/anaconda/conda-bld/work/gcc-4.9.1
++ uname
+ '[' Darwin == Darwin ']'
+ ./configure --prefix=/Users/rwest/anaconda/envs/_build --with-gmp=/Users/rwest/anaconda/envs/_build --with-mpfr=/Users/rwest/anaconda/envs/_build --with-mpc=/Users/rwest/anaconda/envs/_build --with-isl=/Users/rwest/anaconda/envs/_build --with-cloog=/Users/rwest/anaconda/envs/_build --disable-multilib
checking build system type... x86_64-apple-darwin14.4.0
checking host system type... x86_64-apple-darwin14.4.0
checking target system type... x86_64-apple-darwin14.4.0
checking for a BSD-compatible install... /usr/bin/install -c
checking whether ln works... yes
checking whether ln -s works... yes
[...]

PR #401, opened 2015-09-03
gcc-4.8: Fix default include path for system headers

stuarteberg:
There is a critical problem with conda's gcc package, which this PR addresses.

**tl;dr**: If we install gcc to `$PREFIX`, then gcc automatically considers `$PREFIX/include` to be a "system" header path.  That can mess up the include order.  This PR installs gcc to a special place (`$PREFIX/gcc`) to avoid the problem.

**Background:**

As we all know, when gcc searches for a header file, it searches any paths the user provided via `-I`, and then gcc also searches an internal list of "system" directories.  In general, the user's directories are searched in the same order as the user's `-I` arguments. But if the user supplies a `-I` argument that happens to *already* be on gcc's internal "system" include path, then that particular `-I` argument is ignored -- it's a system directory, and system directories are always searched last.  (See gcc docs: https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html)

**The problem:**

When building gcc, the `--prefix` configure argument specifies the installation directory.  As it turns out, gcc **automatically adds** `$(prefix)/include` to its internal "system" include path. [1]  Since conda's gcc recipe `--prefix=$PREFIX`, the environment's `$PREFIX/include` directory is considered to be a *system* directory by gcc.

\[1\]: The [gcc configuration docs](https://gcc.gnu.org/install/configure.html) say:
>Both the local-prefix include directory and the GCC-prefix include directory are part of GCC's “system include” directories.

**Why is that bad?**

Now the user can't control the include order properly.  In my case, an old version of a library (say, hdf5) happened to be installed in `/usr/local/include`.  Ordinarily, it shouldn't cause any problems because I passed `-I $PREFIX/include` to the compiler, so my hdf5 headers in `$PREFIX/include` should be found first and used in my program.  But since gcc thinks `$PREFIX/include` is a system header, my `-I` argument is ignored, and that directory is searched last.  The outdated hdf5 headers in `/usr/local/include` were found first, and my recipe didn't compile properly.

**The Fix:**

gcc can be configured to install its files in custom install locations for `lib`, `bin`, `share`, etc., but apparently `include` cannot be separately controlled.  The only way to change the installation `include` directory is to change the entire install prefix.  In this PR, instead of `--prefix=$PREFIX`, I use `--prefix=$PREFIX/gcc`.  Then, to maintain compatibility with existing recipes, I also provide extra configure args to make sure that `bin`, `lib`, etc. stay in the same locations they had previously.  Only the gcc headers have moved.

**Addendum:**

While I was at it, I fixed some other minor issues with the `gcc` package.  I could have opened separate PRs for those, but in the past it's been easier to consider multiple changes at once (gcc is tedious to compile).

### DIFF
--- a/gcc-4.8/build.sh
+++ b/gcc-4.8/build.sh
@@ -11,7 +11,6 @@ ln -s "$PREFIX/lib" "$PREFIX/lib64"
 
 if [ "$(uname)" == "Darwin" ]; then
     export LDFLAGS="-Wl,-headerpad_max_install_names"
-    export BOOT_LDFLAGS="-Wl,-headerpad_max_install_names"
     export DYLD_FALLBACK_LIBRARY_PATH="$PREFIX/lib"
 
     ./configure \
--- a/gcc-4.8/build.sh
+++ b/gcc-4.8/build.sh
@@ -12,6 +12,7 @@ ln -s "$PREFIX/lib" "$PREFIX/lib64"
 if [ "$(uname)" == "Darwin" ]; then
     export LDFLAGS="-Wl,-headerpad_max_install_names"
     export BOOT_LDFLAGS="-Wl,-headerpad_max_install_names"
+    export DYLD_FALLBACK_LIBRARY_PATH="$PREFIX/lib"
 
     ./configure \
         --prefix="$GCC_PREFIX" \
--- a/gcc-4.8/build.sh
+++ b/gcc-4.8/build.sh
@@ -10,8 +10,10 @@ mkdir "$GCC_PREFIX"
 ln -s "$PREFIX/lib" "$PREFIX/lib64"
 
 if [ "$(uname)" == "Darwin" ]; then
-    export LDFLAGS="-Wl,-headerpad_max_install_names"
-    export DYLD_FALLBACK_LIBRARY_PATH="$PREFIX/lib"
+    # On Mac, we expect that the user has installed the xcode command-line utilities (via the 'xcode-select' command).
+    # The system's libstdc++.6.dylib will be located in /usr/lib, and we need to help the gcc build find it.
+    export LDFLAGS="-Wl,-headerpad_max_install_names -Wl,-L${PREFIX}/lib -Wl,-L/usr/lib"
+    export DYLD_FALLBACK_LIBRARY_PATH="$PREFIX/lib:/usr/lib"
 
     ./configure \
         --prefix="$GCC_PREFIX" \
@@ -24,8 +26,8 @@ if [ "$(uname)" == "Darwin" ]; then
         --with-mpc="$PREFIX" \
         --with-isl="$PREFIX" \
         --with-cloog="$PREFIX" \
-        --with-boot-ldflags=$LDFLAGS \
-        --with-stage1-ldflags=$LDFLAGS \
+        --with-boot-ldflags="$LDFLAGS" \
+        --with-stage1-ldflags="$LDFLAGS" \
         --enable-checking=release \
         --with-tune=generic \
         --disable-multilib
--- a/gcc-4.8/build.sh
+++ b/gcc-4.8/build.sh
@@ -1,3 +1,12 @@
+# Install gcc to its very own prefix.
+# GCC must not be installed to the same prefix as the environment,
+# because $GCC_PREFIX/include is automatically considered to be a
+# "system" header path.
+# That could cause -I$PREFIX/include to be essentially ignored in users' recipes
+# (It would still be on the search path, but it would be in the wrong position in the search order.)
+GCC_PREFIX="$PREFIX/gcc"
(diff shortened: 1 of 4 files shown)


stuarteberg:
BTW, You can get a pre-compiled version of this PR with my channel:

    conda install -c stuarteberg gcc=4.8.5

stuarteberg:
PS -- The easiest way to check the gcc include order is with `gcc -v`:

With the old gcc package (comments added):
```
$ echo | gcc -v -x c -
...
...
#include "..." search starts here:
#include <...> search starts here:
 /miniconda/envs/myenv/lib/gcc/x86_64-unknown-linux-gnu/4.8.5/include
 /usr/local/include
 /miniconda/envs/myenv/include #<-- *** Whoops, that's not supposed to be a system path ***
 /miniconda/envs/myenv/lib/gcc/x86_64-unknown-linux-gnu/4.8.5/include-fixed
 /usr/include
End of search list.
...
```

With this PR's gcc package (comments added):
```
$ echo | gcc -v -x c -
...
...
#include "..." search starts here:
#include <...> search starts here:
 /miniconda/envs/myenv/lib/gcc/x86_64-unknown-linux-gnu/4.8.5/include
 /usr/local/include
 /miniconda/envs/myenv/gcc/include #<-- *** It's okay if this is a system path ***
 /miniconda/envs/myenv/lib/gcc/x86_64-unknown-linux-gnu/4.8.5/include-fixed
 /usr/include
End of search list.
...
```

asmeurer:
Is this ready to merge and test?

stuarteberg:
>Is this ready to merge and test? 
I've done some testing, but now I'm building a couple larger packages just to be sure (e.g. boost).  I'll know in a few minutes if there were any surprises.

BTW, I had trouble testing this on OS X because of #302, and a few other issues related to the linker.  I had to make a few additional commits (see the commit log).  Without those commits, I don't understand how you got the original package to compile on OSX.  Clearly I might be missing something.

Anyway, the current recipe settings seem to work just fine, but I would appreciate a quick review.  (It should be easier to grok if you review the commits one at a time.)

stuarteberg:
OK, I can build boost-1.55.0 without problems on CentOS 5.11, Ubuntu 14.04, and OS X 10.10.  (I did not test any 32-bit systems.)

Meanwhile, shellcheck found one tiny quote issue, which I've now pushed (via a rebase).

stuarteberg:
This PR is ready, as far as I know.

asmeurer:
OK, I'll try building it and let you know if I come across any issues..

stuarteberg:
>OK, I'll try building it and let you know if I come across any issues.. 

@asmeurer, have you gotten a chance to test the new gcc package yet?  Any idea when you'll be pushing your [new build](https://binstar.org/asmeurer/gcc/4.8.5/download/linux-32/gcc-4.8.5-2.tar.bz2) over to the "official" anaconda default channel?

stuarteberg:
Just a quick ping to the participants of #279: The gcc recipe was updated again last week, via this PR.  If you copied the gcc package to your own channel, you should eventually upgrade to the new version.  BTW @frol, I assume these changes should also be made to the `gcc-5` recipe, too.

At the moment, the new version is not yet pushed to the default anaconda channel, but for now you can use @asmeurer's build: `conda install -c asmeurer gcc=4.8.5`.  (The new version has `build_number=2`.)

@jakirkham, @connie, @orodbhen, @mjuric, @frol, @izaid, @dfroger

frol:
@asmeurer BTW, my PR #373 with `make strip-install` hasn't still made into the master. Is there any (design?) problem?

@stuarteberg Sure I will update `gcc-5` recipe. While we are here again, I have revisited post-link and pre-unlink scripts and the logic of `if [ "$(uname)" == "Darwin" ]; then exit ; fi` (https://github.com/stuarteberg/conda-recipes/blob/ed718bb5f44f53e63a7947842b60244dac1ab6db/gcc-4.8/post-link.sh#L3) seems wrong-targeted to me. IMHO, it would be better to use `if [ "$(uname)" != "Linux" ]; then exit ; fi`

stuarteberg:
>I have revisited post-link and pre-unlink scripts and the logic of `if [ "$(uname)" == "Darwin" ]; then exit ; fi` seems wrong-targeted to me. IMHO, it would be better to use `if [ "$(uname)" != "Linux" ]; then exit ; fi`

The logic in `post-link.sh` could (in theory) be applied to any unix-ish OS, including Darwin.  But we happen to know that Darwin doesn't need the fix, so we figured it would be better just to skip post-link entirely in that case.  (It wouldn't do any harm to run it on Darwin, but it seems faster/safer just to skip post-link since we know it isn't needed.)

But someday when conda supports other non-Linux OS's (I dunno... Solaris?), then we probably wouldn't want to skip post-link.  For now, a special case for Darwin is fine.

PR #402, opened 2015-09-03
gsl - use system compiler instead of Anaconda gcc.

pavoljuhas:
gsl package should either require libgcc in its run dependencies or not use the Anaconda gcc at all.


msarahan:
Incorporated in #564

PR #405, opened 2015-09-04
Make pandoc recipe work for Windows

jschueller:


jschueller:
@ccordoba12, hi, what about this one too ?

ccordoba12:
Sure!! Great addition too!

Thanks a lot for your help with these packages :-)

PR #410, opened 2015-09-08
Update cmake to 3.3.1

jschueller:


cmake/meta.yaml
```diff
@@ -3,9 +3,9 @@ package:
   version: 3.2.2
```
thisch: You forgot to change this line

jschueller:
oups

ccordoba12:
Thanks!

PR #411, opened 2015-09-08
LAPACK should not require Python

jschueller:


jakirkham:
Probably, could be reworded to be clearer. LAPACK shouldn't require Python.

jschueller:
@jakirkham, ok

jakirkham:
Moved into https://github.com/conda/conda-recipes/pull/482

PR #417, opened 2015-09-12
Added python library search path to osx to enable boost-python compilation on OS X with Python 3

inati:


frol:
It is wierd. It was me, who added Python 3 support to boost conda package and it compiled fine on Linux and Mac OS Python 2 and 3 with gcc-5 and clang.

BTW, I don't see any problem in setting `--with-python-root` explicitly. I'm just wondering why it worked for me.

Looks good to me!

ccordoba12:
I'll take a look at this one next week :-)

inati:
@ccordoba12 have you had a chance to look at this yet?

ccordoba12:
Merging. I'll test it on my side when building the packages on OS X :-)

PR #422, opened 2015-09-16
Add clink 0.4.5 (windows only)

tadeu:


msarahan:
Sorry you closed this - did we not get to it fast enough?  Please ping me on Windows issues like this.  I use Cmder, which uses Clink and find it helpful.  I'd be happy to include it here.

clink/activate_clink.bat
```diff
@@ -0,0 +1,2 @@
+@echo off
+call clink inject -q --profile "%LOCALAPPDATA%"\clink
```
msarahan: Is there a way to make sure that clink is removed cleanly when this package is uninstalled?

clink/bld.bat
```diff
@@ -0,0 +1,6 @@
+@mkdir %SCRIPTS% 2> NUL
+@for %%f in (clink.bat clink.lua clink_inputrc_base clink_x64.exe clink_dll_x64.dll clink_x86.exe clink_dll_x86.dll) do (
+    move %SRC_DIR%\%%f %SCRIPTS%
+    @if errorlevel 1 exit 1
+)
+xcopy %RECIPE_DIR%\activate_clink.bat %PREFIX%\etc\conda\activate.d\
```
msarahan: newline here?  I think it runs fine without it, but I like to keep them present.

clink/bld.bat
```diff
@@ -0,0 +1,6 @@
+@mkdir %SCRIPTS% 2> NUL
+@for %%f in (clink.bat clink.lua clink_inputrc_base clink_x64.exe clink_dll_x64.dll clink_x86.exe clink_dll_x86.dll) do (
```
msarahan: You don't copy the license.  Does that make the package non-GPL compliant?

tadeu:
@msarahan I closed it because I didn't think it was useful: at least for me, I abandoned `clink` in favor of using `cmder` (which already combines `ConEmu`, `clink` and other useful things).

PR #426, opened 2015-09-25
Add a new recipe for miktex

jschueller:
allows pdf conversion on windows for nbconvert


ccordoba12:
Very interesting addition!

PR #428, opened 2015-09-28
Add support for building python-bioformats

jakirkham:


jakirkham:
After playing around with this, I have decided this is not the best want to install these packages at present. May revisit in the future.

PR #430, opened 2015-09-30
Add hy

jakirkham:
A Lisp in Python. <https://github.com/hylang/hy>


kalefranz:
Probably should remove all of the comments from these stubs before we merge them in here.

jakirkham:
How does that look @kalefranz?

PR #432, opened 2015-10-01
SimpleITK 0.9.1 update

blowekamp:
Adding more specific library dependencies to avoid missing libraries at runtime. There appears to be no check for matching library so version between build time and run time if not manually specified.

### DIFF
--- a/simpleitk/bld.bat
+++ b/simpleitk/bld.bat
@@ -1,8 +1,12 @@
 mkdir C:\b
 cd C:\b
 
-if "%ARCH%" == "32" set "CMAKE_GENERATOR=Visual Studio 9 2008"
-if "%ARCH%" == "64" set "CMAKE_GENERATOR=Visual Studio 9 2008 Win64"
+set "CMAKE_GENERATOR=Visual Studio 9 2008"
+if "%PY_VER%" == "3.3" set "CMAKE_GENERATOR=Visual Studio 10"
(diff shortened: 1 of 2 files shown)


simpleitk/bld.bat
```diff
@@ -1,8 +1,12 @@
 mkdir C:\b
 cd C:\b
 
-if "%ARCH%" == "32" set "CMAKE_GENERATOR=Visual Studio 9 2008"
-if "%ARCH%" == "64" set "CMAKE_GENERATOR=Visual Studio 9 2008 Win64"
+set "CMAKE_GENERATOR=Visual Studio 9 2008"
+if "%PY_VER%" == "3.3" set "CMAKE_GENERATOR=Visual Studio 10"
+if "%PY_VER%" == "3.4" set "CMAKE_GENERATOR=Visual Studio 10"
+if "%PY_VER%" == "3.5" set "CMAKE_GENERATOR=Visual Studio 13"
```
ccordoba12: These are

* `Visual Studio 10 2010`
* `Visual Studio 14 2015`

according to this page

https://cmake.org/cmake/help/v3.3/manual/cmake-generators.7.html

blowekamp: Thanks. You are correct  with the Visual Studio 14.

The years are optional for recent versions of Visual Studio.

ccordoba12:
Thanks for the fix :-)

blowekamp:
Thank you for the prompt merge!!

PR #439, opened 2015-10-13
add console_shortcut

msarahan:
Adds a simple recipe that installs a Windows start menu shortcut to start a command prompt with a particular environment activated. The environment that is activated is the one that console_shortcut was installed into.

This is part of a 4-part series of PRs to add the capability of installing shortcuts to environments other than root. Other PRs:

conda/conda#1699
conda/conda-env#178
ContinuumIO/menuinst#4


msarahan:
This PR duplicates https://github.com/ContinuumIO/anaconda/pull/268

I don't see a way around this right now.  This package will replace anaconda.bat for Windows, and is thus required in the Anaconda repository.  However, it really also needs to be public.  Thus, I am duplicating it.

PR #440, opened 2015-10-19
Separate test data from matplotlib package

mdboom:
matplotlib includes almost 36MB of compressed test data, and most users won't need it.  This separates out matplotlib from the test data, in a new package called "matplotlib-tests".

This is my first real foray into conda packaging, so I may not have done some things correctly.

The upcoming matplotlib 1.5 release will behave normally if the test data is not installed (you just can't run the tests).  I've set this to git URLs for now, but obviously I'll change that to a regular old tarball once a release has been made and before this is merged.  I'd just like to get feedback on this approach in advance of the matplotlib release in case this will require further changes on matplotlib's end.

### DIFF
--- a/matplotlib-tests/bld.bat
+++ b/matplotlib-tests/bld.bat
@@ -1,1 +1,1 @@
-xcopy lib\matplotlib\tests %SP_DIR%\matplotlib /S
+move lib\matplotlib\tests %SP_DIR%\matplotlib\
--- a/matplotlib-tests/bld.bat
+++ b/matplotlib-tests/bld.bat
@@ -1,0 +1,1 @@
+xcopy lib\matplotlib\tests %SP_DIR%\matplotlib /S
--- a/matplotlib-tests/meta.yaml
+++ b/matplotlib-tests/meta.yaml
@@ -1,0 +1,18 @@
+package:
+  name: matplotlib-tests
+  version: 1.5.0
+
+source:
+  git_url: https://github.com/matplotlib/matplotlib.git
+  git_rev: v1.5.x
+
+requirements:
+  build:
+    - matplotlib ==1.5.0
+
+  run:
+    - matplotlib ==1.5.0
+
+about:
+  home: http://matplotlib.sourceforge.net/
+  license: PSF-based (http://matplotlib.sourceforge.net/users/license.html)
--- a/matplotlib-tests/meta.yaml
+++ b/matplotlib-tests/meta.yaml
@@ -15,6 +15,15 @@ requirements:
   run:
     - matplotlib =={{ version }}
 
+  test:
+  requires:
+    - nose
+    - mock
+    - matplotlib
+
+  imports:
+    - matplotlib
+
 about:
   home: http://matplotlib.sourceforge.net/
   license: PSF-based (http://matplotlib.sourceforge.net/users/license.html)
--- a/matplotlib-tests/meta.yaml
+++ b/matplotlib-tests/meta.yaml
@@ -1,6 +1,8 @@
+{% set version = '1.5.0' %}
+
 package:
   name: matplotlib-tests
-  version: 1.5.0
+  version: {{ version }}
 
 source:
   git_url: https://github.com/matplotlib/matplotlib.git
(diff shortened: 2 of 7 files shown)


matplotlib-tests/bld.bat
```diff
@@ -0,0 +1 @@
+xcopy lib\matplotlib\tests %SP_DIR%\matplotlib /S
```
msarahan: Have you tested this?  I only ask because I ALWAYS get xcopy syntax wrong, and end up copying all files to the given folder, rather than the folder itself.

I find it more intuitive to use robocopy, or move for single folders:

    move lib\matplotlib\tests %SP_DIR%\matplotlib\

Move might be faster, too, but probably not really something to optimize for.

matplotlib-tests/meta.yaml
```diff
@@ -0,0 +1,18 @@
...
+  name: matplotlib-tests
+  version: 1.5.0
+
+source:
+  git_url: https://github.com/matplotlib/matplotlib.git
+  git_rev: v1.5.x
+
+requirements:
+  build:
+    - matplotlib ==1.5.0
```
msarahan: I don't like pinning to exact versions.  I get your point here, but is there a way that might require less maintenance in the future?

matplotlib/cfg_notests.patch
```diff
@@ -0,0 +1,13 @@
+diff --git setup.cfg.template setup.cfg.template
```
msarahan: This patch is not referenced in meta.yaml, and thus will not be applied.  Did you mean to include it?

matplotlib-tests/bld.bat
mdboom: Sorry, I should have mentioned I haven't tested this.  I'll have to fire up my Windows partition (unless there's some automatic testing here I can take advantage of).

matplotlib-tests/meta.yaml
mdboom: Is there a way to get the version of the current package (on line 3 above) and use that?  Really, the versions have to be the same for this all to work.

matplotlib/cfg_notests.patch
msarahan: scratch that, the newline between source and patch sections threw me.

mdboom: I think it is referenced in meta.yaml (on line 12) -- or am I missing something?

mdboom: Ah.  No worries.

matplotlib-tests/meta.yaml
msarahan: Might be something like this: https://github.com/conda/conda-build/pull/8

I haven't used it, and don't see it used much, but I like the idea.

mdboom: Interesting.  I've updated this to use jinja to avoid the version number duplication.  It's still not perfect, since now the package version is defined in a non-standard way (not just in the `package.version` YAML value)...

msarahan:
I will test this on Windows sometime soon and let you know how it works.  Looks great to me otherwise.

mdboom:
Thanks for your help.  I'll come back and update this to point to real tarballs once the matplotlib 1.5.0 release is out.

msarahan:
I'd keep the source as a git source.  I prefer them generally, except from the point of view of verifiability (if anyone on the other end deletes and recreates a tag or rev, then we have little detection capability.  I don't know if we have any internal policy to this kind of decision, but we probably should for the sake of security.

The upside of git repositories is that for debugging recipes.  When a recipe is failing, you're left with a _build env and a work folder in conda-bld.  Since that work folder is a git repo, any changes you make to source are very easily added to patches:

git diff --no-prefix > patch_filename

Any changes you make to patches are sort of cumulative: just overwrite the patch with the current one, and the changes from the last one carry over, since they were applied after checking out the source.

If you have a tarball instead, then you have to manually maintain a separate reference version of the source, and do all the git init and git add stuff.

mdboom:
@msarahan: That's good to know.  I'm fine either way -- whatever the best practice is from those with a lot of conda packaging experience than myself.  But at a minimum I'll change it from a branch (`v1.5.x` in this case) to a tag (`v1.5.0` when there is one..).  ;)

mdboom:
This is updated to use the final matplotlib 1.5.0 release and is ready to merge IMHO.

msarahan:
Thanks!

PR #442, opened 2015-10-23
Pin `pylibtiff`'s version of `libtiff` to 4.0.3 or less

jakirkham:
`pylibtiff` doesn't currently have support for a release of `libtiff` later that 4.0.3. So, pin it such that it doesn't use a later release.

### DIFF
--- a/pylibtiff/meta.yaml
+++ b/pylibtiff/meta.yaml
@@ -12,15 +12,15 @@ source:
 requirements:
   build:
     - python
-    - numpy
+    - numpy 1.9.*
     - libtiff <=4.0.3
   run:
     - python
-    - numpy
+    - numpy 1.9.*
     - libtiff <=4.0.3
 
 build:
-  number: 2
+  number: 3
 
 test:
   requires:
--- a/pylibtiff/meta.yaml
+++ b/pylibtiff/meta.yaml
@@ -12,15 +12,15 @@ source:
 requirements:
   build:
     - python
-    - numpy 1.9.*
+    - numpy
     - libtiff <=4.0.3
   run:
     - python
-    - numpy 1.9.*
+    - numpy
     - libtiff <=4.0.3
 
 build:
-  number: 3
+  number: 2
 
 test:
   requires:
--- a/pylibtiff/meta.yaml
+++ b/pylibtiff/meta.yaml
@@ -13,14 +13,14 @@ requirements:
   build:
     - python
     - numpy
-    - libtiff
+    - libtiff <=4.0.3
   run:
     - python
     - numpy
-    - libtiff
+    - libtiff <=4.0.3
 
 build:
-  number: 1
+  number: 2
 
 test:
   requires:


msarahan:
Uh oh, this will cause problems.  With the new builds all being vc-featured on windows, we'll either need to build older libtiff versions, or fix whatever the problem is with pylibtiff.  I'd prefer the latter.

jakirkham:
Unfortunately, the author of `pylibtiff` has been unresponsive of late so that may be tricky.

msarahan:
OK.  I can't make this a high priority, since it is not a default Anaconda package.  If you come up with a fix PR, I'd be happy to look at it.  If the author does not include it, I can at least include your patch in building pylibtiff for the conda repo.

jakirkham:
Well, using the library with this newer version of `libtiff` fails on import so it is kind of nasty. Could we at least pin it?

jakirkham:
Well, using the library with this newer version of `libtiff` fails on import so it is kind of nasty. This is a problem on all platforms.

Could we at least pin it? That will at least allow it to build on some platforms.

I can't promise when I would get to that, if at all. This sort of change ends up being a bit more complicated than you might realize. It requires a rather complex config file ( e.g. https://github.com/pearu/pylibtiff/blob/master/libtiff/tiff_h_4_0_3.py ) be added for each version. This even varies on supposed bugfix releases.

If you really are concerned about Windows, maybe an updated copy of the `libtiff` recipe could be included here. This could serve double duty in terms of both helping build this and provide an example of how to do build with the new Windows recipe.

msarahan:
You can change the recipe however you want.  Here's what I'd like you to also factor in, though:

  - Continuum can't possibly promise backwards support of old library versions.  We will not support recompiling old versions of libtiff with features.
  - Old builds do not use features, so their VC runtime dependencies are not well-defined.  For non-Windows platforms, it doesn't matter.  For Windows, this means that you have a high probability of having a mismatched runtime dependency.  Your program will likely fail to run anyway, and if it doesn't, then it will probably have interesting bugs related to mixing runtimes.

If you are willing to support builds of the older libtiff that have VC features associated, then your fix (version pinning) is a good one.  Otherwise, I'm afraid it only leaves people with more subtle issues at best.

msarahan:
Also, just for my edification: what does pylibtiff provide that more mainstream libraries like pillow do not?

jakirkham:
> Continuum can't possibly promise backwards support of old library versions.

Fine. I'm not asking for that. I would at least like to see this work now on 2/3 instead of 0/3 platforms.

> Old builds do not use features, so their VC runtime dependencies are not well-defined.

Understood. This seems like a beneficial change in the long run. Admittedly, a breaking one though.

> Your program will likely fail to run anyway, and if it doesn't, then it will probably have interesting bugs related to mixing runtimes.

This is where I think there is some confusion. I don't maintain `pylibtiff`. I have, within the past few months, written one patch. When it comes to supporting a new version of `libtiff`, I feel I have already demonstrated that this is non-trivial and non-obvious. Addressing this would require some active participation from the author to get this done.

> We will not support recompiling old versions of libtiff with features.

As far as fixing this, I will reiterate myself and ask that you please supply the current `libtiff` recipe(s) back here and it can be set for a lower version.

jakirkham:
> Also, just for my edification: what does pylibtiff provide that more mainstream libraries like pillow do not?

It supplies a Python library that works directly with TIFFs. However, you could have guessed that.

The TIFF format has often been jokingly referred to as a "Thousand Image File Formats". Initially, it began as a black/white format and has since evolved to include grayscale and color. TIFFs have grown to allow things like pages, chunking, compression, etc. Also, they initially had size limits that have been overcome by BigTIFF. Some of these features are non-standard or are implemented in ways where they may break with some of the TIFF standard. They also provide a variety of options for metadata that have grown over the years.

As a result, many libraries like `pillow` do not support all variants of TIFF either because it simply hasn't been prioritized or it would be challenging to implement. There are other alternatives like `tifffile`, despite how awesome it is, does not supports some types of TIFFs and it lacks some metadata support. What is great about `pylibtiff` is it uses memory mapping of TIFFs, supports a wide variety of the TIFF formats, and is designed as a lower level interface to `libtiff`.

msarahan:
Cool.  Thanks for the explanation of why pylibtiff is a good, necessary thing.  I will update the libtiff recipe on conda-recipes shortly.

msarahan:
https://github.com/conda/conda-recipes/pull/445

jakirkham:
Thanks @msarahan. I'll take a closer look at it and let you know if I have questions.

jakirkham:
Alright, I have managed to get `tifffile` to work for my needs. So, I am abandoning and closing this patch.

jakirkham:
I have added this PR ( https://github.com/conda/conda-recipes/pull/534 ) to remove `pylibtiff` as it can't be supported.

PR #445, opened 2015-10-27
Update libtiff

msarahan:
Sorry this includes console_shortcut too.  I tried to cherry-pick, but failed.  I have no time to do a clean git commit right now.  Leave this here until I have time, if that is important.

ping @groutr


stuarteberg:
I've cherry-picked f430134159d24dd78406e9cdeaa8286de3063925 from here into #462, so you should close PR and just move ahead with that one.

PR #451, opened 2015-10-29
add netifaces

minrk:
used by jupyter-client. This is a simple conda-skeleton-pypi package, onlg ensuring that it is installed with pip instead of setup.py, since it unconditionally imports setuptools.

How do we get things like this into official conda packages?

### DIFF
--- a/netifaces/bld.bat
+++ b/netifaces/bld.bat
@@ -1,0 +1,8 @@
+"%PYTHON%" -m pip install --no-use-wheel --ignore-installed .
+if errorlevel 1 exit 1
+
+:: Add more build steps here, if they are necessary.
+
+:: See
+:: http://docs.continuum.io/conda/build.html
+:: for a list of environment variables that are set during the build process.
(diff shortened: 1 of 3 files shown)


ccordoba12:
Pinging @msarahan and @groutr about it.

msarahan:
Happy to include this.  What version of jupyter-client requires this?  I'm asking because I want to know if this is something I need to rush to get into the Anaconda 2.4 release.

ccordoba12:
According to the PR referenced above, it's optional for now but it'll be a requirement since jupyter-client 4.2.

msarahan:
thanks.  I'll put this on my "after release" list.

minrk:
No version yet has been chosen as a hard dependency, but 4.2 at the earliest, possibly holding it to 5.0, neither of which is expected particularly soon. How reliable upstream packaging is is part of what we are considering in deciding if/when we can make it a dependency, since it is a C extension. It's always been 'recommended', though, since IPython 2, I believe.

msarahan:
Great, we'll do what we can to make it readily available in Anaconda.  Hopefully you have good luck on the wheel side, too.

msarahan:
Thanks!

PR #453, opened 2015-10-29
Fix VTK package on OS X

milljm:
For those of us only using the Command Line Tools (CLT).

It would appear conda arbitrarily setting MACOSX_DEPLOYMENT_TARGET to 10.6, assumes one has the Xcode SDK installed.

I was able to determine; There is only one place where Command Line Tools can be installed. And if the user's environment indeed has their development path set to the CLT path... then lets drop the SDK requirement.

The included patch file is to make application bundles work on OS X. See VTK's explination here: http://www.vtk.org/Wiki/Cocoa_VTK#Use_Application_Bundles

### DIFF
--- a/vtk/apple.patch
+++ b/vtk/apple.patch
@@ -1,0 +1,10 @@
+diff -ur a/CMakeLists.txt b/CMakeLists.txt
+--- CMakeLists.txt	2015-09-01 15:41:26.000000000 -0600
++++ CMakeLists.txt	2015-10-20 14:46:28.000000000 -0600
+@@ -599,3 +599,6 @@
+       "without manually building the 'VTKData' target.")
+   endif()
+ endif()
++if(APPLE)
++  set(EXECUTABLE_FLAG MACOSX_BUNDLE)
++endif(APPLE)
--- a/vtk/meta.yaml
+++ b/vtk/meta.yaml
@@ -7,9 +7,10 @@ source:
   fn:  vtk-6.3.0.tar.gz
   patches:
     - dlls.patch                              [win]
+    - apple.patch                             [osx]
 
 build:
-  number: 1                                   [linux]
+  number: 1                                   [not win]
   has_prefix_files:                           [win]
     - Lib/site-packages/vtk/__init__.py       [win]
 


milljm:
Forgot to mention... That apple.patch file... This may not be the best place for it... as this is an upstream issue...   Your call =D

aeslaughter:
This patch, specifically, the Apple bundle portion is needed for VTK python bindings to work properly on OSX. Without this change you cannot interact with the rendered VTK image.

ccordoba12:
I don't understand. Are you trying to make an app bundle using our VTK package?

aeslaughter:
I don't know what an app bundle is, but I know that without this patch VTK interaction does not work correctly on OSX (see http://www.vtk.org/Wiki/Cocoa_VTK#Use_Application_Bundles)

    To be able to display a graphical interface (that is, display windows that allow mouse and keyboard 
    interaction), OS X requires you to use an application bundle. If a plain UNIX executable tries, it will 
    mostly work, but there will be various bugs, such as keyboard and mouse events not working properly. 
    Many, but not all, of the VTK examples are built as plain UNIX executables, and thus have these 
    problems. This is bug 2025. When you build your own VTK application, it is best to make it in the form 
    of an application bundle.

vtk/build.sh
```diff
@@ -32,6 +32,9 @@ if [ `uname` == Darwin ]; then
     CC=cc
     CXX=c++
     PY_LIB="libpython${PY_VER}.dylib"
+    if [ `xcode-select -p` = "/Library/Developer/CommandLineTools" ]; then
+      unset MACOSX_DEPLOYMENT_TARGET
+    fi
```
ccordoba12: It's not conda which assumes this, it's cmake. I had to dig a lot to find the 10.6 SDK (so I can post it somewhere if people needs it) but we need it to ensure that VTK runs from 10.6 onwards, not from the system version VTK was built in.

So, I don't accept this change. Of course, you can just fork our recipe and add this line if you want, but I'd like to ask you to remove it here.

ccordoba12:
There are two unrelated changes here:

1. The first is the addition of `apple.patch` which is necessary to use VTK from the command line. I'm very grateful for that because I didn't know about it.
2. Trying to unset `MACOSX_DEPLOYMENT_TARGET` to avoid installing the 10.6 SDK. I don't agree with it because it wouldn't allow us to ensure that VTK can be used from 10.6 onwards.

Once change 2. is reverted, I'll merge this and create new VTK packages.

@milljm, thanks a lot for your contribution, by the way :-)

aeslaughter:
Related to what @milljm is seeing, I just  did the following on El Capitan with XCode 7.1 installed:

```
conda install conda-build
git clone git@github.com:conda/conda-recipes.git
cd conda-recipes
conda build vtk
```
This fails with:
```
~/projects/conda-recipes (master)$ conda build vtk
Error: The build field should be a dict, not unicode
```

This is easily fixed by changing line 12 of `meta.yaml` to use "not win".

After making this change, I again run `conda build vtk` and get the following error:
```
(master)$ conda build vtk
Removing old build environment
Removing old work directory
BUILD START: vtk-6.3.0-1
Solving package specifications: ...........
The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    cmake-3.3.1                |                0        17.2 MB

The following NEW packages will be INSTALLED:

    cmake:      3.3.1-0      
    openssl:    1.0.2d-0     
    pip:        7.1.2-py27_0 
    python:     2.7.10-2     
    readline:   6.2-2        
    setuptools: 18.4-py27_0  
    sqlite:     3.8.4.1-1    
    tk:         8.5.18-0     
    wheel:      0.26.0-py27_1
    zlib:       1.2.8-0      

Fetching packages ...
cmake-3.3.1-0. 100% |#####################################################################################################################################| Time: 0:00:19 932.44 kB/s
Extracting packages ...
[      COMPLETE      ]|########################################################################################################################################################| 100%
Linking packages ...
[      COMPLETE      ]|########################################################################################################################################################| 100%
Removing old work directory
Source cache directory is: /Users/slauae/conda-bld/src_cache
Downloading source to cache: vtk-6.3.0.tar.gz
Downloading http://www.vtk.org/files/release/6.3/VTK-6.3.0.tar.gz
vtk-6.3.0.tar. 100% |#####################################################################################################################################| Time: 0:00:01  24.71 MB/s
Success
Extracting download
Package: vtk-6.3.0-1
source tree in: /Users/slauae/conda-bld/work/VTK-6.3.0
+ mkdir build
+ cd build
++ uname
+ '[' Darwin == Linux ']'
++ uname
+ '[' Darwin == Darwin ']'
+ CC=cc
+ CXX=c++
+ PY_LIB=libpython2.7.dylib
+ cmake .. -DCMAKE_C_COMPILER=cc -DCMAKE_CXX_COMPILER=c++ -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/Users/slauae/.conda/envs/_build -DCMAKE_INSTALL_RPATH:STRING=/Users/slauae/.conda/envs/_build/lib -DBUILD_DOCUMENTATION=OFF -DVTK_HAS_FEENABLEEXCEPT=OFF -DBUILD_TESTING=OFF -DBUILD_EXAMPLES=OFF -DBUILD_SHARED_LIBS=ON -DVTK_WRAP_PYTHON=ON -DPYTHON_EXECUTABLE=/Users/slauae/.conda/envs/_build/bin/python -DPYTHON_INCLUDE_PATH=/Users/slauae/.conda/envs/_build/include/python2.7 -DPYTHON_LIBRARY=/Users/slauae/.conda/envs/_build/lib/libpython2.7.dylib -DVTK_INSTALL_PYTHON_MODULE_DIR=/Users/slauae/.conda/envs/_build/lib/python2.7/site-packages -DVTK_USE_OFFSCREEN=ON -DModule_vtkRenderingMatplotlib=ON -DVTK_USE_X=OFF
-- The C compiler identification is AppleClang 7.0.0.7000176
-- The CXX compiler identification is GNU 5.2.0
CMake Error at /Users/slauae/.conda/envs/_build/share/cmake-3.3/Modules/Platform/Darwin.cmake:76 (message):
  CMAKE_OSX_DEPLOYMENT_TARGET is '10.6' but CMAKE_OSX_SYSROOT:

   ""

  is not set to a MacOSX SDK with a recognized version.  Either set
  CMAKE_OSX_SYSROOT to a valid SDK or set CMAKE_OSX_DEPLOYMENT_TARGET to
  empty.
Call Stack (most recent call first):
  /Users/slauae/.conda/envs/_build/share/cmake-3.3/Modules/CMakeSystemSpecificInformation.cmake:36 (include)
  CMakeLists.txt:28 (project)


-- Configuring incomplete, errors occurred!
See also "/Users/slauae/conda-bld/work/VTK-6.3.0/build/CMakeFiles/CMakeOutput.log".
Command failed: /bin/bash -x -e /Users/slauae/projects/conda-recipes/vtk/build.sh
```

aeslaughter:
FYI, I also have CommandLineTools installed.

milljm:
Having Command Line Tools installed seems to be what kills the build. Unfortunately, to even clone conda-recipes, one needs to install Command Line Tools in order to obtain the `git` binary.... rats!

It looks as though, we can determine a correct SDK path based on the results of xcrun:

Xcode only installed:
```bash
> xcrun --show-sdk-platform-path
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform
```

Xcode Command Line Tools installed:
```bash
> xcrun --show-sdk-platform-path
xcrun: error: unable to lookup item 'PlatformPath' from command line tools installation
xcrun: error: unable to lookup item 'PlatformPath' in SDK '/'

> xcrun --show-sdk-version
10.11
```

Perhaps we can 'do something' based on the combined results of xcode-select and xcrun. I'll try to figure out a smarter way!

Also, (just a helpful FYI), I figured Conda was setting the SDK version arbitrarily to 10.6 based on Conda's Documentation: http://conda.pydata.org/docs/building/environment-vars.html 
I had no idea it was Cmake doing it :) . Sorry!

ccordoba12:
> Having Command Line Tools installed seems to be what kills the build

No, I'm also have command line tools installed, and I can assure that's not the problem. The problem is you don't have the 10.6 SDK installed, as I said above. If I understand things correctly, an SDK is a package that contains the core libraries used for a particular Mac version (like 10.6, 10.11, etc). 

> Also, (just a helpful FYI), I figured Conda was setting the SDK version arbitrarily to 10.6 based on Conda's Documentation: http://conda.pydata.org/docs/building/environment-vars.html
I had no idea it was Cmake doing it :) . Sorry!

Conda sets `MACOSX_DEPLOYMENT_TARGET` to assure that all our Mac packages can be ran from a particular version onwards (10.6 for now), so it's not arbitrary.

However, it is CMake the one that demands that the SDK that corresponds to  `MACOSX_DEPLOYMENT_TARGET` is installed in the build machine, not Conda. It's true, as you discovered, that by unsettling that variable you can make CMake work, but that will make the generated VTK package dependent on the system version it was built.

----

I hope this is clearer @milljm and @aeslaughter. In short, if you don't have the 10.6 SDK package installed, you can't compile VTK, sorry, but that's how CMake works.

milljm:
Ahhh I see. Yeah, I didn't understand that what I was changing, would make this only work for my machine build. I forget that these recipes build the very binary `conda install` is downloading. My bad!

Thanks for clearing that up! I'll modify my pull request to satisfy your requirements shortly.

ccordoba12:
Thanks a lot @milljm!

milljm:
@ccordoba12  Done!
Thanks again for your help!

ccordoba12:
No problem :-)

Again, thanks a lot for your help too!

aeslaughter:
When does this become available with: conda install vtk=6.3.0?

Thanks for implementing this fix.

ccordoba12:
I have to recompile the package and make it available in our official channels. I think it'll be ready this weekend or early next week.

PR #454, opened 2015-10-30
Add pysparselm and sparselm packages

hovren:
pysparselm is a Python wrapper for the sparselm library.
Sparselm solves sparse non-linear least-squares problems.

The packages have been tested on Fedora 22 x64.
The sparselm package requires that the "suitesparse" package is installed.

### DIFF
--- a/pysparselm/build.sh
+++ b/pysparselm/build.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/sh
+
+sed -i -e 's/fpic/fPIC/' setup.py
+export CC="gcc -fPIC"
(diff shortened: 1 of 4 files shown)


msarahan:
Where can people get the suitesparse package?  Is that a Fedora system package?  Is there any way to build and package it here?  I ask because it's always nice when recipes don't explicitly depend on some other package manager being used to provide dependencies.

hovren:
It appears to be from here, according to the RPM package in Fedora: http://faculty.cse.tamu.edu/davis/suitesparse.html

It does look like a package that might be useful or even used by other packages already, but I could not find it in conda already.
I have not tried building it.

PR #455, opened 2015-10-30
aubio/meta.yaml: depend on numpy version

piem:
This adds x.x to numpy so that package gets build against current CONDA_NPY version.


msarahan:
thanks!

PR #457, opened 2015-11-01
Add Git for Windows and Nano recipes (both windows only recipes)

msarahan:
This is intended for usage at Software Carpentry workshops.  The idea is to be able to bootstrap a complete installation by only installing miniconda, then installing a conda environment: https://github.com/swcarpentry/swc-conda-recipes/pull/1

That PR has builds at https://anaconda.org/swc that were built using the recipes contained in this PR.


ccordoba12:
You also have a recipe for `console_shortcut` here, but you just merged another PR with the same(?) recipe.

Please fix :-)

ccordoba12:
By the way, great to have `nano` on Windows! We should also add recipes for clink and ConsoleZ too ;-)

msarahan:
I'm not aware of ConsoleZ - I need to learn about it.  I have been intending to get Cmder (my favorite) into a recipe.  It rolls in clink.  Maybe better to have them separate, though?

msarahan:
I will try to clean up console_shortcut, but cleaning up merge commits generally does not end well for me.

ccordoba12:
I tried Cmder but it failed for me because i haven't the VS 2015 redistributable (I think). ConsoleZ is simpler, but it doesn't have that requirement.

msarahan:
Git for windows has a small issue wherein git bash is not getting Library\bin for the environment on its PATH.  Needs a fix here.

console_shortcut/menu-windows.json
```diff
@@ -0,0 +1,12 @@
+{
+    "menu_name": "Python Prompts",
+    "menu_items":
+    [
+        {
+            "name": "Conda Env Prompt",
```
JanSchulz: Shouldn't have this the env name?

msarahan: Actually, that recipe shouldn't be in this PR at all.  Thanks for pointing it out.  I'll try to clean it up.

PR #458, opened 2015-11-01
update git to 2.6.2; build without tcltk (no GUI)

msarahan:
TclTk was causing problems with Mac packaging.  I think it was a dumb path escaping issue: the Git GUI app name has a space in it.

This also forces the build to use our builds of zlib and openssl, rather than any system versions.  The -m64 flag might need special-casing in an ARCH statement.

### DIFF
--- a/git/build.sh
+++ b/git/build.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 make configure
-./configure --prefix=$PREFIX
+./configure --prefix=$PREFIX --without-tcltk
 make all
 make install
 
--- a/git/build.sh
+++ b/git/build.sh
@@ -1,7 +1,13 @@
 #!/bin/bash
 
+# NOTE: On Ubuntu, compilation required installation of autoconf and gettext from apt-get
+
 make configure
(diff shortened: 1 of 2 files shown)


ccordoba12:
Does this fix pulling from `https` repos? Last time I tried this recipe, it wasn't working (and I couldn't fix it)

msarahan:
Tell me more about this issue, please.  I'm not aware of such things - but I have been using Git primarily on Windows, and then on Mac only the system one that came pre-installed.

FWIW, I did clone an https URL with this built binary, and it worked fine...

msarahan:
also, note that this links against our versions of openssl and zlib.  That might have been the issue before?

ccordoba12:
Well, if you don't have write permissions to a repo, you can only clone it using the `https` protocol, like for example:

https://github.com/conda/conda-recipes.git

It's also useful if you don't want to be setting ssh keys in all the machines you use :-) That's why I tend to prefer it over ssh.

But the git that comes up from this recipe can't handle `https` for some reason :-s

msarahan:
I think it does now.  By all means, try it!

https://anaconda.org/swc/git

ccordoba12:
Will do! I meant the previous recipe by the way ;-)

msarahan:
Closing.  Superseded by https://github.com/conda/conda-recipes/pull/492

PR #459, opened 2015-11-02
Add functools32

jakirkham:
A backport package of `functools` with the `lru_cache` to python 2.7.

### DIFF
--- a/functools32/meta.yaml
+++ b/functools32/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: functools32
-  version: "3.2.3-2"
+  version: "3.2.3.2"
 
 source:
   fn: functools32-3.2.3-2.tar.gz
--- a/functools32/meta.yaml
+++ b/functools32/meta.yaml
@@ -9,10 +9,10 @@ source:
 
 requirements:
   build:
-    - python >=2.7,<3
+    - python 2.7.*
 
   run:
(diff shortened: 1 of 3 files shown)


msarahan:
Please clean up the comments from conda-skeleton.  Also, since this is specific to Python 2.7, should you pin a version dependency on Python at runtime to <3, so that this is never installed in a Python 3 environment?  Does that even make sense?

jakirkham:
Sorry about that. Cleaned up the comments.

Also, the version has been pinned to be some form of 2.7 as it doesn't support 2.6 or earlier either.

functools32/meta.yaml
```diff
@@ -0,0 +1,24 @@
...
+  fn: functools32-3.2.3-2.tar.gz
+  url: https://pypi.python.org/packages/source/f/functools32/functools32-3.2.3-2.tar.gz
+  md5: 09f24ffd9af9f6cd0f63cb9f4e23d4b2
+
+requirements:
+  build:
+    - python >=2.7,<3
+
+  run:
+    - python >=2.7,<3
```
ccordoba12: This can simplified by using

    2.7*

jakirkham: Done.

ccordoba12:
It looks good to me. Merging.

jakirkham:
Thanks.

PR #461, opened 2015-11-02
aniso8601

faph:
requirement for flask-restful

### DIFF
--- a/aniso8601/bld.bat
+++ b/aniso8601/bld.bat
@@ -1,0 +1,2 @@
+"%PYTHON%" setup.py install
+if errorlevel 1 exit 1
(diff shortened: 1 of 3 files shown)


aniso8601/bld.bat
```diff
@@ -0,0 +1,2 @@
+"%PYTHON%" setup.py install --single-version-externally-managed --record=record.txt
```
msarahan: Many of your recipe PRs include this.  I think you're using it to get rid of the egg?  That's fine, but on Windows, I have never gotten ```--single-version-externally-managed``` to work without the ```--root=c:\``` argument also being provided.  I generally use ```--old-and-unmanageable``` instead, which seems to behave equivalently to the two-argument solution, just with a slightly unnerving name.

If you have tested this on Windows, I'll take your word for it and merge it.

faph:
Yes, picked this up somewhere on the mailing list to get rid of eggs. The '--record' option is required to get it to work.

I've tested this on Windows, yes. Not on anything else though.

msarahan:
Thanks!

PR #466, opened 2015-11-06
Fixed failing build of SimpleITK on osx.

zivy:
Cmake flags are propagated to the external project builds in the
SuperBuild. The install (CMAKE_INSTALL_PREFIX) flag for ITK
caused the build to fail. This was removed as we are not doing any
install. Also turned the BUILD_EXAMPLES  flag to off (default is on),
as these are not required for building SimpleITK.


ccordoba12:
+1 to what @stuarteberg said.

zivy:
As we are not installing any of the toolkits in the superbuild for SimpleITK there is no real reason for setting the value of this flag, which is why I removed it.

ccordoba12:
@zivy, I'm trusting on you with this one :-)

PR #474, opened 2015-11-10
Add recipe for TensorFlow

jjhelmus:


tensorflow/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - python
+    - numpy >=1.9.2
+    - six >=1.10.0
+
+test:
+  imports:
+    - tensorflow
+
+about:
+  home: http://tensorflow.com/
```
msarahan: URL seems to be dead for me

msarahan: Maybe https://github.com/tensorflow/tensorflow would be better?

jjhelmus: Interesting, that is the URL that Google is putting in the Wheel metadata but it is not correct.  Should be http://tensorflow.org/.  I'll update momentary.

tensorflow/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# install using pip from the whl file provided by Google
+
+if [ `uname` == Darwin ]; then
+    pip install https://storage.googleapis.com/tensorflow/mac/tensorflow-0.5.0-py2-none-any.whl
+fi
+
+if [[ (`uname` == Linux) ]]; then
+    pip install https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-0.5.0-cp27-none-linux_x86_64.whl
```
msarahan: This is OK, but why not build from source?  Looks like it would require one other package (Bazel): https://github.com/tensorflow/tensorflow/blob/master/tensorflow/g3doc/get_started/os_setup.md#detailed_install

jjhelmus:
Bazel is a pain to install.  It requires a JVM install and and only supports select versions of Ubunutu, [install documentation](http://bazel.io/docs/install.html).  I'm not interested in building TensorFlow from source, I'm happy with the wheel files.

msarahan:
OK.  That's fine for now.  I'll investigate building Bazel when I have time.  Maintaining the ability to build from source is extremely important for long-term packaging.  Thanks for this contribution.

groutr:
I've been building TensorFlow this morning, and I can confirm that getting all the prereqs is an annoyance.  Once everything is in place, the build process is almost completely automatic, requiring the user to answer only a single question at config time.  A tensorflow package from source would be pretty tricky.  The easiest way would probably build a pip package from source, and then install it in the build environment.

ahmadia:
Thanks for this.  I'm going to try these recipes out and upload them to our memex channel.

ahmadia:
@msarahan - I'll ping the Google developers once these are uploaded, but do you think my time is better spent trying to build these from source or looking at a Windows package?

msarahan:
I think building from source will be necessary for obtaining a Windows package.

ahmadia:
Ahh, good point.

PR #477, opened 2015-11-13
Update yt to 3.2.2

ngoldbaum:
This is a regularly scheduled bugfix release. We'd appreciate it if Continuum could build packages for python2.7 and python3.4. Python3.5 support should also be doable but we don't have CI set up for that yet so it might be premature to set up conda packages.

Thanks for your help distributing yt!

### DIFF
--- a/yt/meta.yaml
+++ b/yt/meta.yaml
@@ -1,11 +1,11 @@
 package:
   name: yt
-  version: 3.2.1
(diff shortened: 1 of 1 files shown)


ccordoba12:
Thanks. I added an entry to our weekly meeting notes to remember to do that for next week :-)

PR #479, opened 2015-11-19
ipyparallel: Add recipe to build 4.1.0.

jakirkham:
Related https://github.com/ContinuumIO/anaconda-recipes/issues/18

### DIFF
--- a/ipyparallel/meta.yaml
+++ b/ipyparallel/meta.yaml
@@ -35,8 +35,6 @@ requirements:
     - ipython >=4
     - jupyter_client
     - ipykernel
-    - nose
-    - notebook
     - tornado >=4
 
   run:
@@ -47,8 +45,6 @@ requirements:
     - ipython >=4
     - jupyter_client
     - ipykernel
-    - nose
-    - notebook
     - tornado >=4
 
 test:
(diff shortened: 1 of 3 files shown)


ipyparallel/meta.yaml
```diff
@@ -0,0 +1,84 @@
...
+    - python
+    - ipython_genutils
+    - decorator
+    - pyzmq >=13
+    - ipython >=4
+    - jupyter_client
+    - ipykernel
+    - nose
+    - notebook
+    - tornado >=4
```
ccordoba12: Please remove the last three dependencies, i.e. nose, notebook and tornado. They are not needed to build or run this package.

jakirkham:
Actually, it was unable to build without those.

ccordoba12:
Please take a look here: https://github.com/ipython/ipyparallel/blob/master/setup.py#L85

The only addition for 4.1 seems to be tornado, not nose or the notebook.

jakirkham:
Oh, that was suppose to be another line note. Sorry, posted on my phone.

Yes, to use everything in `ipyparallel` they are. ( https://github.com/ipython/ipyparallel/issues/73 ) That being said, these are python and some web langs mixed in, which really isn't so much of a burden to install.

ccordoba12:
But the idea of the Jupyter split is to have a very clean installation for each of its packages, not to mix packages that are not needed for runnning or building other components.

So please remove those deps. You can install them separately :-)

jakirkham:
Corrected.

ccordoba12:
Thanks! :-)

PR #481, opened 2015-11-21
Add glpk recipe (both windows and unix)

aebrahim:


glpk/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+export CFLAGS="-fPIC -O3"
+./configure --prefix=$PREFIX --with-pic
```
ccordoba12: Why `fpic` and `pic`?

aebrahim: It's there because I have had errors with old versions not linking
correctly with distutils. I'll test without it to see if that's still the
case.

aebrahim:
OK, I removed the PIC stuff. It works fine without it on both Mac/Linux.

glpk/meta.yaml
```diff
@@ -0,0 +1,23 @@
...
+source:
+  fn: glpk-4.57.tar.gz
+  url: http://ftp.gnu.org/gnu/glpk/glpk-4.57.tar.gz
+  md5: 237531a54f73155842f8defe51aedb0f
+
+build:
+  features:
+    - vc9      [win and py27]
+    - vc10     [win and py34]
+    - vc14     [win and py35]
```
ccordoba12: Does this package have Python bindings?

ccordoba12: I mean, if not, you can select just one `vc` feature here (for example, `vc14`) instead of listing them all :-)

aebrahim: Not inherently... But I have a python package than links against it and I
know there are others as well. And if you build it with vc14 then it can't
be used with python 2.7

ccordoba12: Ok, no problem then :-)

glpk/bld.bat
```diff
@@ -0,0 +1,14 @@
+cd w%ARCH%
+copy config_VC config.h
+
+nmake /f Makefile_VC
+
+mkdir %PREFIX%\bin
```
Juanlu001: We're having problems with this recipe on Windows and I think it's because it's copying the binary to the wrong location. See http://conda.pydata.org/docs/building/environment-vars.html for the value of `LIBRARY_BIN` and https://github.com/snorfalorpagus/conda-recipes/blob/116af80/glpk/bld.bat#L5 for a recipe that works.

Juanlu001: Original issue: https://github.com/CAChemE/pyomo-recipes/issues/6

msarahan: Thanks for reporting this. If you have time, please submit a PR so that you
get credit for the fix.

Juanlu001: Besides @snorfalorpagus published his own glpk recipe five days ago and includes some testing, so perhaps he wants to weigh in. Otherwise I'm just going to edit `bld.bat` as little as possible.

msarahan: Please add the testing, if possible.  That would be very nice to have.

snorfalorpagus:
@Juanlu001 I think you are correct, that on Windows `%LIBRARY_PREFIX%` should be used instead of `%PREFIX%`, or the more specific variables (`%LIBRARY_BIN%`, etc).

My build is a bit of a hack, as it downloads a pre-compiled version and just installs it. This is largely owing to my lack of knowledge on compiling things on the Windows stack.

The test I added is quite simple and uses an `.mps` file from the GLPK examples folder. It figured it's better than nothing!

This recipe doesn't seem to copy the `.dll` shared library, which should live in the `%LIBRARY_BIN%` folder - it's not enough just to copy the static library, as someone might have a need to link dynamically.

Juanlu001:
Perhaps we can combine the best of both worlds :) I'll submit my changes this evening.

PR #482, opened 2015-11-23
Update lapack to 3.6.0

jschueller:


lapack/build.sh
```diff
@@ -2,10 +2,11 @@ mkdir build
...
 
 cmake \
-    -DCMAKE_INSTALL_PREFIX=$PREFIX \
-    -DBUILD_TESTING=OFF \
-    -DBUILD_SHARED_LIBS=ON \
-    ..
+  -DCMAKE_INSTALL_PREFIX=$PREFIX \
+  -DBUILD_TESTING=OFF \
+  -DBUILD_SHARED_LIBS=ON \
+  -DLAPACKE=ON \
```
ccordoba12: What is this new option used for?

jakirkham: I believe this is for the C interface to LAPACK. Am a bit surprised this wasn't already enabled.

jschueller:
@ccordoba12 
This enables lapacke, the c bindings for lapack, see http://www.netlib.org/lapack/lapacke.html.
The next OpenTURNS version is able to use it instead of needing a fortran compiler, so I enabled it.

jakirkham:
Includes https://github.com/conda/conda-recipes/pull/411

jschueller:
@jakirkham, unfortunately the previous 3.5.0 version had no cmake config files and so it could not be automagically detected out of the box, so we did not use it.

ccordoba12:
Ok, thanks for the clarifications :-) Merging!

PR #485, opened 2015-11-25
[WIP] build_helper.py for making .binstar.yml files and submitting

PeterDSteinberg:
This is a work in progress.  It is a ```build_helper.py``` script for adding the anaconda-build's required ```.binstar.yml``` file to each recipe and submitting those packages to anaconda.org.

#### See comment below about changes in CLI. 
<strike>An example of writing all the ```.binstar.yml``` files for the r-packages packages:

python build_helper.py psteinberg/abc r-packages/*  psteinberg --exception-action raise  --channel dev --anaconda-build-platforms osx-64 --submit
</strike>
The command above says to 
 * edit the ```.binstar.yml``` files in all packages in ```r-packages```
 * make ```psteinberg``` the anaconda upload user
 * ```--submit``` the packages after writing ```.binstar.yml```
 * use the ```psteinberg/abc``` build queue for submittal
 * raise any exceptions along the way
 * upload to the ```dev``` channel
 * build only on platform ```osx-64```

The catch is that we need to run our own build workers so that we don't overwhelm the public queue or internal continuum ones with tons of packages.  Here are the steps for running a build worker (See also [docs.anaconda.org](http://docs.anaconda.org/):
 * Install anaconda-build and anaconda-client based on the instructions in their repos
 * ```anaconda worker register USERNAME/QUEUE``` replacing USERNAME/QUEUE with a queue name of that form you have created under your username
* The register command above should print out a worker_id to use in the following command:
* ```anaconda worker run WORKER_ID --cwd /some/dir/for/builds ```
* Make sure you are submitting packages to the queue you have created.

TODO: Things to fix:
 * atom-shell has a meta.yaml that is not yaml.loading okay
 * some meta.yaml files have Jinja template variables that have not been interpolated correctly yet

### DIFF
--- a/build_helper.py
+++ b/build_helper.py
@@ -9,6 +9,8 @@ import time
 import traceback
 import yaml
 
+from conda_build.metadata import MetaData
+
 
 DEFAULT_ANACONDA_ARCH = [
       'linux-32',
@@ -85,7 +87,8 @@ def submit_package(args, top_dir, package_name):
 
 def on_each_package(args, top_dir, package_name):
     if args.action == 'edit':
-        meta_file = join(top_dir, package_name, 'meta.yaml')
+        package_path = join(top_dir, package_name)
+        meta_file = join(package_path, 'meta.yaml')
         if not exists(meta_file):
             has_subdirs = False
             for try_levels in range(1,10):
@@ -100,8 +103,7 @@ def on_each_package(args, top_dir, package_name):
                 raise ValueError(msg)
             else:
                 return
-        with open(meta_file, 'r') as f:
-            meta = yaml.load(f.read())
+        meta = MetaData(package_path).meta
         binstar_yml_file = join(top_dir, package_name, '.binstar.yml')
         if args.yaml_load_existing and exists(binstar_yml_file):
             with open(binstar_yml_file, 'r') as f:
--- a/build_helper.py
+++ b/build_helper.py
@@ -1,0 +1,170 @@
+from __future__ import print_function, unicode_literals, division
+
+import argparse
+import glob
+from os.path import exists, join, abspath, split, walk, isdir
+import subprocess as sp
+import sys
+import time
+import yaml
+
+DEFAULT_ANACONDA_ARCH = [
+      'linux-32',
+      'linux-64',
+      'osx-32',
+      'osx-64',
+      'win-32',
+      'win-64',
+]
+
+BINSTAR_YAML_TEMPLATE = {
+         'after_failure': ['echo "after_failure!"'],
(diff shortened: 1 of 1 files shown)


PeterDSteinberg:
The CLI has been cleaned up.  There are two basic forms of using it:
```
python build_helper.py edit ...
```
(edits .binstar.yml)
and
```
python build_helper.py submit
```
(submits to anaconda.org)

A full example: Edit all the binstar.yml files, skipping and printing exceptions, building for osx-64 platform and the psteinberg username.
```
python build_helper.py edit * psteinberg --platforms osx-64 --exception-action stringify
```
And here is an example submitting one package after the command above or similar one has been run to create a binstar.yml in the package:
```
python build_helper.py submit r-packages/r-maps psteinberg/abc --channel dev --exception-action raise
```

PeterDSteinberg:
The commit above makes several changes to be consistent with [a new PR I did for anaconda-build](https://github.com/Anaconda-Server/anaconda-build/pull/147/files#diff-557a02d4bec6d252fafadacf8741c74aR181).  Specifically, now the .binstar.yml files that are written for each package will have an ```install_channels``` entry that will put anaconda channels like ```r```, ```defaults```, or a user specified channel into condarc on the build worker.  With ```r``` in the ```install_channels```, the worker can do ```conda build .``` rather than ```conda build . -c r```.

build_helper.py
```diff
@@ -0,0 +1,195 @@
...
+         'package': 'package',
+         'platform': DEFAULT_ANACONDA_ARCH,
+         'install': 'conda build .',
+         'user': 'user'
+    }
+
+def cmd(args, cwd='.'):
+    proc = sp.Popen(args, stdout=sp.PIPE,
+                    stderr=sp.PIPE, cwd=cwd)
+    while proc.poll is None:
```
groutr: Is there a better way to wait for the process to finish than sleeping?

PeterDSteinberg: Right.  I'll change that. I had it that way originally thinking I would read stdout as it is generated but I think the processes are short enough to where that is not really needed.

build_helper.py
```diff
@@ -0,0 +1,195 @@
...
+                    has_subdirs = True
+
+            if not has_subdirs:
+                msg = "Package {} in top_dir {} has no meta.yaml"
+                msg = msg.format(package_name, top_dir)
+                raise ValueError(msg)
+            else:
+                return
+        with open(meta_file, 'r') as f:
+            meta = yaml.load(f.read())
```
groutr: You should be using conda's loading functions for meta.yaml, since meta.yaml is not standard yaml.

PeterDSteinberg:
This functionality has moved to ContinuumIO/ProtoCI

PR #487, opened 2015-11-25
Improve boost recipe on Linux and Mac

pavoljuhas:
Remove duplicate code and options set to boost build-system defaults.
Use MACOSX_DEPLOYMENT_TARGET from conda build-system environment instead of explicit compiler option.  Avoid using patchelf on Linux and the resulting larger and more fragile binaries.

### DIFF
--- a/boost/meta.yaml
+++ b/boost/meta.yaml
@@ -8,6 +8,7 @@ source:
   md5: 6aa9a5c6a4ca1016edd0ed1178e3cb87
 
 build:
+  binary_relocation: False      [linux]
   features:
     - vc9               [win and py27]
     - vc10              [win and py34]
--- a/boost/meta.yaml
+++ b/boost/meta.yaml
@@ -1,11 +1,11 @@
 package:
   name: boost
-  version: 1.59.0
+  version: 1.61.0
 
 source:
-  fn:  boost_1_59_0.tar.bz2
-  url: http://sourceforge.net/projects/boost/files/boost/1.59.0/boost_1_59_0.tar.bz2
-  md5: 6aa9a5c6a4ca1016edd0ed1178e3cb87
+  fn:  boost_1_61_0.tar.bz2
+  url: http://sourceforge.net/projects/boost/files/boost/1.61.0/boost_1_61_0.tar.bz2
+  md5: 6095876341956f65f9d35939ccea1a9f
 
 build:
   features:
@@ -16,7 +16,7 @@ build:
 requirements:
   build:
     - python
-    - icu               [unix]
+    - icu 54.*          [unix]
     - bzip2             [unix]
     - zlib
 
@@ -24,7 +24,7 @@ requirements:
     # python dependency is here due to libboost-python library that depends on
     # python version
     - python
-    - icu               [unix]
+    - icu 54.*          [unix]
     - zlib
 
 about:
--- a/boost/meta.yaml
+++ b/boost/meta.yaml
@@ -30,4 +30,3 @@ requirements:
 about:
   home: http://www.boost.org/
   license: Boost-1.0
-
--- a/boost/meta.yaml
+++ b/boost/meta.yaml
@@ -8,7 +8,6 @@ source:
   md5: 6aa9a5c6a4ca1016edd0ed1178e3cb87
 
 build:
-  binary_relocation: False      [linux]
   features:
     - vc9               [win and py27]
     - vc10              [win and py34]
--- a/boost/bld.bat
+++ b/boost/bld.bat
@@ -22,16 +22,22 @@ if errorlevel 1 exit 1
     address-model=%ARCH% ^
     variant=release ^
     threading=multi ^
-    link=shared ^
+    link=static,shared ^
     -j%CPU_COUNT% ^
     -s ZLIB_INCLUDE="%LIBRARY_INC%" ^
     -s ZLIB_LIBPATH="%LIBRARY_LIB%"
 if errorlevel 1 exit 1
 
+for /F "tokens=1,2,3 delims=." %%a in ("%PKG_VERSION%") do (
+   set PKG_VERSION_MAJOR=%%a
+   set PKG_VERSION_MINOR=%%b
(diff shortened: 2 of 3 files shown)


ccordoba12:
Sorry, I think I'm going to decline this pull request. The current one works correctly on Linux and Mac and has been the work of many people.

Your work, for example, ignores the `--with-python-root` option, without which bindings for Python 3 are not built at all.

ccordoba12:
Another example: you ignore the `-stdlib=libstdc++` compiler flag, which was asked to be set by another user in our mailing list so that all our C++ packages are linked againts that library instead of `libc++`.

I could continue but I don't see the point of it.

ccordoba12:
> Avoid using patchelf on Linux and the resulting larger and more fragile binaries

This is the only interesting bit. Why the binaries are more fragile with our current recipe?

ccordoba12:
@pavoljuhas, sorry if I was a bit harsh, but I've worked a lot on the Boost recipe so I feel a bit uncomfortable with your changes :-)

However, I'd really to hear why you think the current binaries are fragile because I'm about to update Boost on our official channels. So your input would be really valuable right now :-)

pavoljuhas:
@ccordoba12, OK, no problem.  I will comment on the other parts shortly, but here is an example of a problem I ran into when building with standard binary relocation:
pavoljuhas/conda-recipes@66c44956ad4f0eebe09a3070d96abf22663a1af8 vs pavoljuhas/conda-recipes@764c66e69a3ec8393446c7a0ec403eb5aa75b5f7 (both in the libobjcryst-patchelf-problem branch).

Both commits have only 2 recipes for the libobjcryst C++ library and for its Python interface pyobjcryst.  The pyobjcryst package uses libobjcryst shared library and boost_python so the sequence to build pyobjcryst is as follows:
```shell
conda build libobjcryst
conda build pyobjcryst
```
In the first commit the libobjcryst conda package is built with a standard binary relocation, and the `conda build` on Linux uses patchelf to adjust the RPATH in libObjCryst.so.  The pyobjcryst package in the second step compiles, but it does not pass the test.  If you activate the `_test` environment, `python -c 'import pyobjcryst'` fails with a segmentation fault.

This problem is avoided when libobjcryst gets built without binary relocation (the second commit).  libObjCryst does not refer to any other Anaconda binaries so this is not a problem.  The pyobjcryst recipe remains the same, but this time it works.  BTW, in both cases the compiled _pyobjcryst.so module gets relocated and has its RPATH set to `$ORIGIN/../../../..`.

pavoljuhas:
Further update to the rpath-patchelf issue: I see this problem on Ubuntu 10.04.4 with g++ 4.4.3.
However, the same recipes work fine on the later Ubuntu 12.04.5 with g++ 4.6.3.

I am not sure how this behaves on CentOS 5, which to my knowledge is used for standard Anaconda packages.  I would prefer to build on an older system, because packages built on Ubuntu 12 are not usable on Ubuntu 10 (GLIBC incompatibility).

ccordoba12:
@pavoljuhas, thanks for the update. I'm using CentOS 5 with gcc 4.4, so it seems we need to use your solution instead of letting conda use patchelf.

How do you want to proceed about this? Do you want to create a new PR adding only your solution on top of the current recipe? I'd really prefer that, then in another PR we could discuss about simplifying the current recipe.

What do you think?

pavoljuhas:
@ccordoba12 - Can you take a look again?

I have restored the `--with-python-root` option, but it is now using `python-config` to resolve headers path.

The `-stdlib=libstdc++` option is to my understanding the default behavior of clang++, so it can be as well left out.  I did not see any difference in the produced binaries (Xcode 7.1.1) after removing that option.

The `-mmacosx-version-min` option has the same effect as `MACOSX_DEPLOYMENT_TARGET` environment variable from `conda-build`.  I think it is better to use the environment variable, because build script will keep working should the deployment target change in the future.  Other removed options are `debug-symbols=off` which is implied by `variant=release` option in b2 and `runtime-link=shared` implied by `link=shared`.

I am not sure about the `-L${LIBRARY_PATH}` in the linkflags (introduced in 4f0672ad3e78e27aeadc9d5fa344443490bbb255), but the icu libraries seems to get detected and linked without it as well.

boost/meta.yaml
```diff
@@ -8,6 +8,7 @@ source:
   md5: 6aa9a5c6a4ca1016edd0ed1178e3cb87
 
 build:
+  binary_relocation: False      [linux]
```
frol: Why binary relocation should be disabled on Linux?

frol: I'm sorry, I haven't read all the comments to the PR before asking. However, I still don't get how you apply `RPATH` to the binaries then. Does it magically get applied by your old gcc?

I've used boost package intensively in my packages and, in fact, I have added Python 3 support to this boost conda-package. I haven't noticed any problems with it on CentOS 6.7 and Ubuntu 14.04. If it is a bug in conda-build, then it should be fixed instead of applying this quite a weird workaround.

pavoljuhas:
@frol - the RPATH is set by adding `linkflags="'-Wl,-rpath,\$ORIGIN'"` to the `b2` options on Linux, this was added in 6a88696a9f51d55c0ca626135ca1035c12834b38.  

I have done more testing and found the [patchelf problem for libObjCryst](https://github.com/conda/conda-recipes/pull/487#issuecomment-161215887) was caused by building with `scons`.  scons does not propagate shell environment variables unless told so.  On Linux `conda-build` sets LD_RUN_PATH which gcc uses for RPATH when linking the shared library, and the `patchelf` utility then changes RPATH that already exists.  In my case scons did not propagate LD_RUN_PATH, and so patchelf worked with a binary that had no RPATH setting.  This apparently caused problems with older gcc-s, but all worked fine once I propagated the LD_RUN_PATH variable through scons.

In short, I don't see a reason anymore for setting the RPATH with linker flags.  I am going to revert 6a88696a9f51d55c0ca626135ca1035c12834b38 and go back to standard binary relocation in `conda-build`.

PR #491, opened 2015-12-03
expat: new recipe (git dependency)

frol:
Expat is an XML parser library written in C. It is a stream-oriented parser in which an application registers handlers for things the parser might find in the XML document (like start tags).

You can test this package by installing it from `salford_systems` channel:

```
$ conda install -c salford_systems expat
```

Package on Anaconda.org: https://anaconda.org/salford_systems/expat

### DIFF
--- a/expat/meta.yaml
+++ b/expat/meta.yaml
@@ -1,0 +1,20 @@
+package:
+  name: expat
+  version: 2.1.0
+
+source:
+  fn: expat-2.1.0.tar.gz
+  url: http://sourceforge.net/projects/expat/files/expat/2.1.0/expat-2.1.0.tar.gz
+  md5: dd7dab7a5fea97d2a6a43f511449b7cd
+
+requirements:
+  build:
+    - gcc
+
+about:
+  home: http://expat.sourceforge.net
+  license: MIT
+  summary: |
+    Expat is an XML parser library written in C. It is a stream-oriented parser
+    in which an application registers handlers for things the parser might find
+    in the XML document (like start tags).
--- a/expat/meta.yaml
+++ b/expat/meta.yaml
@@ -7,10 +7,6 @@ source:
   url: http://sourceforge.net/projects/expat/files/expat/2.1.0/expat-2.1.0.tar.gz
   md5: dd7dab7a5fea97d2a6a43f511449b7cd
 
-requirements:
-  build:
-    - gcc
-
 about:
   home: http://expat.sourceforge.net
   license: MIT
--- a/expat/build.sh
+++ b/expat/build.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+./configure --prefix="$PREFIX"
+
+make
+make install


expat/meta.yaml
```diff
@@ -0,0 +1,20 @@
...
+  version: 2.1.0
+
+source:
+  fn: expat-2.1.0.tar.gz
+  url: http://sourceforge.net/projects/expat/files/expat/2.1.0/expat-2.1.0.tar.gz
+  md5: dd7dab7a5fea97d2a6a43f511449b7cd
+
+requirements:
+  build:
+    - gcc
```
msarahan: Is GCC a strict requirement, or can it be compiled with, for example, clang?  Any tips on why GCC might be an explicit requirement?

frol: @msarahan I haven't tried clang. I have successfully built this package for linux-32, linux-64, and osx-64 as it is now. I just prefer to use as much conda dependecies as possible instead of relying on system-wide packages.

ccordoba12: Missing `run:` with `libgcc`, right?

expat/meta.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+about:
+  home: http://expat.sourceforge.net
+  license: MIT
+  summary: |
```
ccordoba12: Why this `|` here?

expat/meta.yaml
frol: @ccordoba12 You may add it there if you like, but it is not needed there.

expat/meta.yaml
frol: I use it to make a multiline string: http://stackoverflow.com/questions/3790454/in-yaml-how-do-i-break-a-string-over-multiple-lines

It seems that `>` is better here, though.

expat/meta.yaml
ccordoba12: I don't feel comfortable about not adding it. It has been our custom to always add `libgcc` when we demand `gcc`. Or not use `gcc` at all (as Mike suggested above).

frol: @ccordoba12 I don't get the logic at all. System-wide packages are less reliable than conda dependency, but you suggest to choose between (1) loosing `gcc` dependency or (2) adding the unnecessary `libgcc` dependency for the sake of "custom". Python is built with gcc isn't it? Still, it doesn't require `libgcc`.

ccordoba12: Nop, Python is not built with `gcc`. It's built with the gcc that comes in CentOS 5. We really use `gcc` for very few packages (the R ones mostly). So until now we've demanded `libgcc` as a run dep whenever we use `gcc` as a build dep.

But perhaps that's only necessary when building C++ libraries instead of C ones (because the C standard is much more stable).

@asmeurer, could you clarify when it's necessary to use `libgcc`?

frol:
I have just removed the `gcc` dependency.

ccordoba12:
@msarahan, merging?

msarahan:
Thanks!

PR #492, opened 2015-12-03
git: great refactoring - 8MB package instead of 93MB (!)

frol:
The most significant shrinkage was achived by the `NO_INSTALL_HARDLINKS=1` - 8MB package instead of 93MB is a significant improvement.

I have also disabled perl utils and git-gui as them (from my point of view) are rarely used:

* `NO_TCLTK` disables git-gui
* `NO_PERL` disables all perl-based utils: git-instaweb, gitweb, git-cvsserver, git-svn (http://www.spinics.net/lists/git/msg99803.html)
* `NO_GETTEXT` disables internationalization (localized message translations) - avoids `gettext` build and run dependencies

You can test this package by installing it from `salford_systems` channel:

```
$ conda install -c salford_systems git
```

Package on Anaconda.org: https://anaconda.org/salford_systems/git

Please, also review #491 as it is a git dependecy.

### DIFF
--- a/git/meta.yaml
+++ b/git/meta.yaml
@@ -13,7 +13,6 @@ build:
 
 requirements:
   build:
-    - gcc
     - expat
     - curl
     - openssl
--- a/git/meta.yaml
+++ b/git/meta.yaml
@@ -1,16 +1,26 @@
 package:
   name: git
-  version: 2.4.6
+  version: 2.6.3
 
 source:
-  git_url: git@github.com:git/git.git
-  git_tag: v2.4.6
+  fn: v2.6.3.tar.gz
+  url: https://github.com/git/git/archive/v2.6.3.tar.gz
+  md5: 65f659fb55ba49f17de03d1d9a859a83
+
+build:
+  # git hardcodes paths to external utilities (e.g. curl)
+  detect_binary_files_with_prefix: true
 
 requirements:
   build:
+    - gcc
+    - expat
+    - curl
     - openssl
     - zlib
   run:
+    - expat
+    - curl
     - openssl
     - zlib
 
--- a/git/build.sh
+++ b/git/build.sh
@@ -1,12 +1,18 @@
 #!/bin/bash
 
-make configure
-./configure --prefix=$PREFIX
-make all
-make install
+export C_INCLUDE_PATH="$PREFIX/include"
 
-cd $PREFIX
-rm -rf lib lib64
-rm -rf share/man
-strip bin/* || echo
-strip libexec/git-core/* || echo
+# NO_TCLTK disables git-gui
+# NO_PERL disables all perl-based utils:
+#   git-instaweb, gitweb, git-cvsserver, git-svn
+#   /ref http://www.spinics.net/lists/git/msg99803.html
+# NO_GETTEXT disables internationalization (localized message translations)
+# NO_INSTALL_HARDLINKS uses symlinks which makes the package 85MB slimmer (8MB instead of 93MB!)
+make \
+    --jobs="$CPU_COUNT" \
+    prefix="$PREFIX" \
+    NO_TCLTK=1 \
+    NO_PERL=1 \
+    NO_GETTEXT=1 \
+    NO_INSTALL_HARDLINKS=1 \
+    all strip install


frol:
@ilanschnell I saw that you were getting git recipe into conda-recipes. Could you please review the changes?

git/meta.yaml
```diff
@@ -1,16 +1,26 @@
...
 requirements:
   build:
+    - gcc
+    - expat
+    - curl
     - openssl
     - zlib
   run:
+    - expat
+    - curl
```
ccordoba12: `libgcc` seems to be missing here

ccordoba12:
This is awesome!! Thanks for the improvements @frol! :-)

However, I'd like to have git-gui in our package. How much does it increase its size if we would add it again?

msarahan:
This is great!  Nice work.  See also https://github.com/conda/conda-recipes/pull/458 - is there anything of value in that PR to merge into yours?

git/meta.yaml
frol: @ccordoba12 I've just `ldd`ed and `grep`ed the binaries, and I couldn't find `libgcc` dependency there.

frol:
@ccordoba12 I have no measures on hands about how much `git-gui` will increase the size because Tcl/Tk requires `gettext` and I couldn't make `gettext` link correctly on my build node.

@msarahan Nothing useful there. I also avoid `autoconf` dependency completely by skipping `configure` step.

msarahan:
OK.  I'll go ahead and close that PR.  This one supersedes it from what I can see.

ccordoba12:
Ok, no problem about git-gui then. Too bad (because I prefer GUIs rather than plain text interfaces) but not a killer :-)

msarahan:
@ccordoba12 let's merge this, and make an issue about getting git-gui built sometime later.  I don't think that is a very high priority, but hopefully one of us might have time to work on it at some point.

msarahan:
Before we merge though, check two things that tripped me up on my build:

- does this include git stash? (mine was mysteriously missing)
- does https checkout work? (mine did not on OSX.  Linux seemed to be OK)

git/meta.yaml
ccordoba12: `libgcc` is the name of the package we use to install just the dynamic libraries provided by gcc instead of the full `gcc` package (which is huge).

I think git is done in plain C, so there shouldn't be a problem when building with a new version of gcc and linking against an old one. Nevertheless, I'd prefer (for the sake of safety :-) to add `libgcc` as run dep here.

ccordoba12:
Before merging, I'd prefer to add `libgcc` as a run dep (here and also on the expat recipe).

frol:
My `git` package builds and works fine for me on linux-32 (CentOS 6.6), linux-64 (CentOS 6.6, Ubuntu 14.04), OS X.

There is an issue on OS X though, I cannot clone HTTPS repos:

```
$ git clone https://github.com/conda/conda-recipes.git /tmp/conda-recipes
Cloning into '/tmp/conda-recipes'...
fatal: unable to access 'https://github.com/conda/conda-recipes.git/': SSL certificate problem: unable to get local issuer certificate
```

frol:
@msarahan HTTPS repos on OS X are broken indeed. Does anyone happen to know how to fix this?

`git stash` works fine on both Linux and OS X.

msarahan:
No idea how to fix it, but I don't think it is an improper build on your part.  Just that certs location might need to be set somehow.

http://stackoverflow.com/questions/24611640/curl-60-ssl-certificate-unable-to-get-local-issuer-certificate

frol:
@ccordoba12 Why would you add `libgcc` dependency where it is not used at all?

```bash
(/tmp/qqenv)[frol]$ which git
/tmp/qqenv/bin/git
(/tmp/qqenv)[frol]$ grep -R libgcc /tmp/qqenv/
(/tmp/qqenv)[frol]$
```

msarahan:
alternatively, we could package a cert as a conda package, but I don't really like the idea of maintaining that responsibility.  Way too much can go horribly wrong if it is not diligently managed.

ccordoba12:
@frol, I'd like to do it for safety. I wouldn't like to deal with odd errors because of a mismatch on gcc versions.

Are there dynamic libs created while building this package that git links against to?

frol:
@ccordoba12 `git` package doesn't get any new shared/static libraries, only executables and some helping templates are packaged. Here is the full list of linked libraries on Linux:

```
$ ldd bin/git* libexec/git-core/* | grep -oP '\tlib[^ ]*' | sort -u
        libc.so.6
        libcom_err.so.2
        libcrypt.so.1
        libcrypto.so.1.0.0
        libcrypto.so.10
        libcurl.so.4
        libdl.so.2
        libexpat.so.1
        libfreebl3.so
        libgssapi_krb5.so.2
        libidn.so.11
        libk5crypto.so.3
        libkeyutils.so.1
        libkrb5.so.3
        libkrb5support.so.0
        liblber-2.4.so.2
        libldap-2.4.so.2
        libnspr4.so
        libnss3.so
        libnssutil3.so
        libplc4.so
        libplds4.so
        libpthread.so.0
        libresolv.so.2
        librt.so.1
        libsasl2.so.2
        libselinux.so.1
        libsmime3.so
        libssh2.so.1
        libssl.so.1.0.0
        libssl.so.10
        libssl3.so
        libz.so.1
```

msarahan:
Potentially helpful: http://stackoverflow.com/a/4454754/1170370

It sounds to me, though, like git may already include these in the bin folder?  Perhaps not in source, though, only in binary installers?

I think the answer is that if no cacert.pem exists in bin/ then we probably need to create one, and I hope the link above helps with that.

ccordoba12:
@frol, ok, no problem with `libgcc` then :-)

frol:
Do you guys can test OS X package? (I have only ssh access to OS X build machine)

Here is what I found: http://superuser.com/questions/721778/can-not-clone-any-git-repository-from-github

>  recently (Jul 2014) had a similar issue and found on OS X (10.9.4) that there was a "DigiCert High Assurance EV Root CA" certificate had expired (although I had another unexpired one as well).
>
> 1. Open Keychain Access
> 2. search Certificates for "DigiCert"
> 3. View menu > Show Expired Certificates
>
>I found two certificates named "DigiCert High Assurance EV Root CA", one expiring Nov 2031 and the expired one at July 2014 (a few of days previously). Deleting the expired certificate resolved the issue for me.

msarahan:
Unfortunately, I didn't find any expired certs on my system (OSX 10.10).

IMHO, the cert issue isn't really your responsibility in this PR.  If we add this package to mainline and make it available for all platforms, then we need to ensure that it works out of the box.  However, as a Conda recipe (not a package), it is already OK - I just wish conda recipes had a good way to document this issue and give people a starting point on how to fix the cert issue.

@ccordoba12 I'd like to merge this.  Any objections?

ccordoba12:
@msarahan, first you need to merge the `expat` recipe. We're having a discussion with @frol about adding `libgcc` as a run dep to `expat` (given that he insists on using the `gcc` package instead of system one).

I think we have two options here:

1. Remove `gcc` as a build dep. I don't see the point of it, given that both `expat` and `git` seem to be C projects.
2. Wait to hear from @asmeurer (maybe @groutr or @ilanschnell can also helps us here) on when to use `libgcc` or not.

    In most recipes I've seen, using `gcc` as a build dep implies using `libgcc` as a run dep. However, that's doesn't seem to be necessary in all cases because (as @frol points out) `git` doesn't end up being linked to `libgcc` (and the same appears to be true for `expat`).

    But that will force us to **manually** check (on a recipe per recipe basis) if the package is linked or not against `libgcc`. That's why I think it's easier (and safer?) just to add `libgcc` as a run dep. Or not to demand `gcc`, if it's not required.

frol:
I have just removed the `gcc` dependency.

ccordoba12:
Thanks! @msarahan, merging?

msarahan:
Thanks!

PR #494, opened 2015-12-04
update opencv to version 3; update TBB recipe a bit

msarahan:
NOTE: For Py2.7 on Windows, this omits the opencv_contrib extras.  These use C++11, and can't be built by the VS 2008 compiler.  Those extras are built for all other platforms (newer Windows targets, as well as all targets for Mac/Linux.


patricksnape:
I actually made a PR to opencv_contrib that fixes [Python 2.7 build errors for Visual Studio 2008](https://github.com/Itseez/opencv_contrib/pull/338). In my conda recipe repository [I have the individual patches](https://github.com/menpo/conda-opencv3/tree/master/conda). 

Your recipe is **a lot** better than mine - I could try giving you a PR that builds using OpenCV contrib if you like? The only module that uses too much C++11 and can't be built at all for Python is the [detection based filters](https://github.com/menpo/conda-opencv3/blob/master/conda/detection_based_filter.patch).

I also haven't gotten it working with Python 3.5 yet and I've been meaning to do that!

opencv/bld.bat
```diff
@@ -0,0 +1,113 @@
...
+
+for /f "delims=" %%A in ('%PREFIX%\python -c "import sys; print(sys.version_info.major)"') DO SET PY_MAJOR=%%A
+
+set "OBJ_DETECT=-DBUILD_opencv_objdetect=false"
+
+IF %PY_MAJOR% EQU 3 (GOTO :PY3) else (GOTO :PY2)
+
+:PY3
+    REM Get python minor version by running a short script:
+    for /f "delims=" %%A in ('python -c "import sys; print(sys.version_info.minor)"') DO SET PY_MINOR=%%A
```
patricksnape: Missing `%PREFIX%` -> `%PREFIX%\python -c`

patricksnape:
Also, OpenCV2 and OpenCV3 are not backwards compatible despite sitting in the same namespace - therefore I'm not sure if you want people who are relying on OpenCV2 to upgrade, get OpenCV3 and then find that their code breaks. Tough decision really - I opted to make mine a new package called `opencv3`.

patricksnape:
OK - I've merged in some changes that gets OpenCV 3.1 building - maybe we should update this branch to just build OpenCV 3.1 - particularly as less patches are needed! Let me know and I will make a PR against your branch if you agree.

patricksnape:
Also - I would like to get this in to mainline conda because it is taking up a quarter of my account storage - there are so many shared libraries, each system is around 30Mb!

msarahan:
Excellent - I'm all for your changes.  Go ahead with your PR.  I will do the work to get this moved onto mainline.

PR #507, opened 2015-12-10
add rtools 3.3 recipe

msarahan:
This bundles GCC and other required build tools using the R-provided Rtools installer.  It copies the files into our Library/bin folder structure.

ping @groutr

### DIFF
--- a/r-packages/r-rtools/bld.bat
+++ b/r-packages/r-rtools/bld.bat
@@ -1,0 +1,21 @@
+FOR %%F IN (%SRC_DIR%\*.exe) DO (
+    set FILENAME=%%F
+    goto found_file
+)
+:found_file
+
+%FILENAME% /SILENT /DIR="%TEMP%\rtools"
+if errorlevel 1 exit 1
+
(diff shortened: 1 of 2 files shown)


frol:
@msarahan I don't know much about Rtools, but it seems that the recipes misses a [skip option](http://conda.pydata.org/docs/building/meta-yaml.html#skipping-builds):

```yaml
build:
  skip: True  # [not win]
```

JanSchulz:
I think R versions are [compatible with certain versions](https://cran.r-project.org/bin/windows/Rtools/), so you would need to package the versions for all supported Rs?

On the other hand I had no problem building r packages for for multiple r versions with only one installed rtools.

groutr:
@frol, you are correct.  This should skip on non windows platforms.

groutr:
@msarahan, it looks like the recipe for R will need to be modified for windows so it can look inside the miniconda folder for the R-tools.  Haven't forgotten this, still working on it.

groutr:
Running this recipes required elevated permissions on my machine.  Use innoextract instead (http://constexpr.org/innoextract/)

PR #512, opened 2015-12-11
nlopt: add details

jschueller:


nlopt/build.sh
```diff
@@ -1,3 +1,3 @@
-./configure --prefix=$PREFIX --enable-shared
+./configure --prefix=$PREFIX --enable-shared --without-octave --without-matlab --without-guile --without-python
```
ccordoba12: why `--without-python`? I think we should leave at least that one :-)

jschueller:
@ccordoba12, it seems not consistent with the windows side, moreover it should depend on python runtime to distinguish the python version. As-is it should not work as we should specify the python location to avoid using host configuration.

jschueller:
@ccordoba12, here you go, proper python bindings

ccordoba12:
Thanks a lot!

PR #529, opened 2015-12-21
New memory_profiler recipe

AlbertDeFusco:
Builds were tested on linux and windows Python 3 and Python 2.


ccordoba12:
Looks good to me.

PR #532, opened 2016-01-03
Update rpy2 to version 2.7.6 and fix failed builds by adding include dir to CFLAGS.

johanneskoester:
The current version of rpy2 in the conda-recipes repository does not build. Here, I added readline as a dependency and fixed the CFLAGS such that the compiler finds it and rpy2 builds properly. I recommend to merge and rebuild quickly, because rpy2 is currently unusable with Python 3.5.1 in conda.

### DIFF
--- a/rpy2/build.sh
+++ b/rpy2/build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-$PYTHON setup.py install
+CFLAGS="-I$PREFIX/include" $PYTHON setup.py install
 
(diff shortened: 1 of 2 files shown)


ccordoba12:
Pinging @groutr about this one

kalefranz:
I don't see any issues with the PR.  Merged it.  @groutr should just be aware of it.

johanneskoester:
Thanks!

PR #536, opened 2016-01-07
Selenium chromedriver

jhedrick:
Updated urls for new version, and added support for installing the binary to `$PYTHONPATH$/Scripts` on windows 32-bit.


selenium-chromedriver/bld.bat
```diff
@@ -1,2 +1,2 @@
-python setup.py install
-if errorlevel 1 exit 1
+for /R "%SRC_DIR%" %%f in (*.exe) do copy %%f "%PYTHONPATH%/Scripts"
```
ccordoba12: Please change this line to

    or /R "%SRC_DIR%" %%f in (*.exe) do copy %%f "%LIBRARY_BIN%"

because `%LIBRARY_BIN%` is now on `PATH`, and its the right location for executables that don't come from Python packages.

PR #539, opened 2016-01-08
Posix_ipc for Linux and Mac

AlbertDeFusco:
An attempt was made to get this working in Windows, but the only way to get real Posix shared memory support seems to be by installing Cygwin.


ilanschnell:
Thanks

PR #542, opened 2016-01-14
Kkwiek/recipes

mutirri:


ccordoba12:
@mutirri, this PR adds some packages already present here (like `appdirs` and `simpy`), and also breaks the convention that all Python packages have to go inside the `python` directory (so now a substantial amount of Python packages are again at the root level of this repo). Please fix this mess.

Pinging also @kalefranz and @msarahan because I don't understand why this PR was merged like this, it shouldn't happen again.

ccordoba12:
Sorry, this doesn't seem to be @mutirri's fault because his PR seems to be done before the `python` dir creation.

But then we should be more careful when merging PRs because this just disrupts everybody's work.

ilanschnell:
Sorry, I should have reviewed more carefully before merging.

mutirri:
Hi @ilanschnell @ccordoba12 @kalefranz @msarahan ,

> Sorry, this doesn't seem to be @mutirri's fault because his PR seems to be done
> before the python dir creation.

Yeah, I didn't know that we are changing policy.
The pull request was created some time ago and there was no such requirement.
I fixed the paths here: https://github.com/conda/conda-recipes/pull/641
It would be great to merge this as soon as possible.

> Sorry, I should have reviewed more carefully before merging.

Ilan, still you could overlook invalid paths since this type of change is not conspicuous.

Anyway sorry for the confusion.

PR #552, opened 2016-01-18
Build graphviz with expat.

djothi:


msarahan:
Looks fine, but can you give us a little more explanation?  Does this close any active issues?  Is any modification to the build scripts necessary, or does expat just get detected and used at configure time?

djothi:
I don't know about any issue related to that fix but graphviz needs expat to generate HTML tables (as stated there: http://www.graphviz.org/Download_source.php).

PR #555, opened 2016-01-18
Move all python packages to a new (root-level) python directory

kalefranz:
This is an initial commit to move python packages out of the root directory for this repo.  Please comment if you see anything that obviously doesn't belong...

More long-term, the root will be reserved for package groupings (like what `r-packages` is now), and we'll continue to refactor into logical groupings.


kalefranz:
@ilanschnell, @groutr, @msarahan One of you mind approving and merging?

ilanschnell:
Did you make sure this doesn't break any existing PRs?

kalefranz:
I went through the PRs since September and merged any that easily could be.  Anything left is still somehow controversial and under discussion or in progress.

ccordoba12:
It's not possible to see the contents of this PR through Github, but if there are errors they should be easy to fix.

Besides, I think it's better to merge this now before merge conflicts start to appear.

ccordoba12:
The only recipes that were left are `ipython-notebook` and `ipython-console`. Should we move or remove them?

The thing is those recipes are for the old IPython packages (i.e. before IPython was split into Jupyter).

ccordoba12:
Also `python-patch` was left, but it's too old, so we should probably remove it too.

ilanschnell:
I agree, `python-patch` should be removed.

ccordoba12:
And what about `ipython-{notebook, qtconsole}`?

kalefranz:
I'm fine removing them all if everyone else is

ccordoba12:
Done :-)

PR #557, opened 2016-01-22
nlopt: depend on libgcc

jschueller:


ccordoba12:
Thanks for the fix, merging :-)

PR #559, opened 2016-01-24
Freeimage: fix win32 build.  Cleanup.

msarahan:
@almarklein please look these over.  These are the changes made in the Anaconda repository.

The only major change here is changing how 64-bit architecture is detected.  The way you had it is only for build-time architecture (what arch is the build machine), not what machine is the target environment.  Thus, my issue building was that I was getting 64-bit DLLs no matter what.


almarklein:
Looks good to me, I always found that line a bit worrying :P

PR #564, opened 2016-01-27
Update GNU GSL to v2.1

bjodah:
This commit updates GNU GSL to the latest version.
It also adds a `run_test.sh` file (in addition to the test suite which is also built in `build.sh`)


bjodah:
I just saw https://github.com/conda/conda-recipes/pull/402
I incorporated those changes in the second commit (but with updated URL and checksum obviously).

msarahan:
Excellent, thanks for the test!

PR #570, opened 2016-01-28
Fix compatibility issues for gcc package

### LINKED ISSUE #563: GCC has the wrong libtool paths
In the GCC package, I find that the `la` files are declaring wrong paths. Everywhere where it is `lib/../lib64`, I think it should be just `lib`, because nothing ever gets placed in `lib64`. This causes problems with everything that wants to use libtool to build the gcc dependencies (as could be the case for #499), as it searches for the GCC standard libraries in the wrong path.

````
~/anaconda3/pkgs/gcc-4.8.5-3 $ find -name *.la -print0 | xargs --null -I {} sh -c 'echo {}; grep lib64 {}'
./gcc/libexec/gcc/x86_64-unknown-linux-gnu/4.8.5/liblto_plugin.la
./lib/libmudflapth.la
libdir='/opt/anaconda1anaconda2anaconda3/lib/../lib64'
./lib/gcj-4.8.5-14/libgjsmalsa.la
libdir='/opt/anaconda1anaconda2anaconda3/lib/../lib64/gcj-4.8.5-14'
./lib/gcj-4.8.5-14/libjvm.la
dependency_libs=' -L/root/miniconda3/conda-bld/work/gcc-4.8.5/x86_64-unknown-linux-gnu/libjava/.libs '/opt/anaconda1anaconda2anaconda3/lib/../lib64'/libgcj.la -L/root/miniconda3/conda-bld/work/gcc-4.8.5/x86_64-unknown-linux-gnu/libjava -lpthread -ldl -lrt'
libdir='/opt/anaconda1anaconda2anaconda3/lib/../lib64/gcj-4.8.5-14'
./lib/libgcj-tools.la
dependency_libs=' -L/root/miniconda3/conda-bld/work/gcc-4.8.5/x86_64-unknown-linux-gnu/libjava -lm '/opt/anaconda1anaconda2anaconda3/lib/../lib64'/libgcj.la -lpthread -ldl -lrt'
libdir='/opt/anaconda1anaconda2anaconda3/lib/../lib64'
./lib/libquadmath.la
libdir='/opt/anaconda1anaconda2anaconda3/lib/../lib64'
./lib/libgij.la
dependency_libs=' -L/root/miniconda3/conda-bld/work/gcc-4.8.5/x86_64-unknown-linux-gnu/libjava/.libs '/opt/anaconda1anaconda2anaconda3/lib/../lib64'/libgcj.la -L/root/miniconda3/conda-bld/work/gcc-4.8.5/x86_64-unknown-linux-gnu/libjava -lpthread -ldl -lrt'
libdir='/opt/anaconda1anaconda2anaconda3/lib/../lib64'
./lib/libssp.la
libdir='/opt/anaconda1anaconda2anaconda3/lib/../lib64'
./lib/libtsan.la
dependency_libs=' -lpthread -ldl '/opt/anaconda1anaconda2anaconda3/lib/../lib64'/libstdc++.la -lm'
libdir='/opt/anaconda1anaconda2anaconda3/lib/../lib64'
./lib/libgomp.la
libdir='/opt/anaconda1anaconda2anaconda3/lib/../lib64'
./lib/libssp_nonshared.la
libdir='/opt/anaconda1anaconda2anaconda3/lib/../lib64'
./lib/libasan.la
dependency_libs=' -lpthread -ldl '/opt/anaconda1anaconda2anaconda3/lib/../lib64'/libstdc++.la -lm'
libdir='/opt/anaconda1anaconda2anaconda3/lib/../lib64'
./lib/gcc/x86_64-unknown-linux-gnu/4.8.5/libcaf_single.la
./lib/gcc/x86_64-unknown-linux-gnu/4.8.5/libgfortranbegin.la
./lib/libsupc++.la
libdir='/opt/anaconda1anaconda2anaconda3/lib/../lib64'
./lib/libstdc++.la
libdir='/opt/anaconda1anaconda2anaconda3/lib/../lib64'
./lib/libgcj.la
libdir='/opt/anaconda1anaconda2anaconda3/lib/../lib64'
./lib/libatomic.la
libdir='/opt/anaconda1anaconda2anaconda3/lib/../lib64'
./lib/libitm.la
libdir='/opt/anaconda1anaconda2anaconda3/lib/../lib64'
./lib/libgfortran.la
dependency_libs=' '/opt/anaconda1anaconda2anaconda3/lib/../lib64'/libquadmath.la -lm'
libdir='/opt/anaconda1anaconda2anaconda3/lib/../lib64'
./lib/libmudflap.la
libdir='/opt/anaconda1anaconda2anaconda3/lib/../lib64'
````

A specific example that fails dur to this:

https://github.com/Zaharid/conda-recipes/tree/master/apfel

### PR DESCRIPTION
**Pull request by Zaharid**
A lib64 folder need to be present on the system fot libtool not to
complain (so fixes  #563) and the way the system info is requested is not
universal enough.

### DIFF
--- a/gcc-4.8/build.sh
+++ b/gcc-4.8/build.sh
@@ -58,7 +58,7 @@ make install-strip
 rm "$PREFIX"/lib64
 
 #Fix libtool paths
-find -name "$PREFIX"/*.la -print0 | xargs -0  sed -i 's%lib/../lib64%lib%g'
+find "$PREFIX"  -name '*.la' -print0 | xargs -0  sed -i 's%lib/../lib64%lib%g'
 
 # Link cc to gcc
 (cd "$PREFIX"/bin && ln -s gcc cc)
--- a/gcc-4.8/build.sh
+++ b/gcc-4.8/build.sh
@@ -35,7 +35,7 @@ else
     # For reference during post-link.sh, record some
     # details about the OS this binary was produced with.
     mkdir -p "${PREFIX}/share"
-	# lsb_release can complain about LSB mosules in stderr, so we
+	# lsb_release can complain about LSB modules in stderr, so we
 	# ignore that.
     lsb_release -a 1> "${PREFIX}/share/conda-gcc-build-machine-os-details"
     ./configure \
--- a/gcc-4.8/build.sh
+++ b/gcc-4.8/build.sh
@@ -35,7 +35,9 @@ else
     # For reference during post-link.sh, record some
     # details about the OS this binary was produced with.
     mkdir -p "${PREFIX}/share"
-    cat /etc/*-release > "${PREFIX}/share/conda-gcc-build-machine-os-details"
+	# lsb_release can complain about LSB mosules in stderr, so we
+	# ignore that.
+    lsb_release -a 1> "${PREFIX}/share/conda-gcc-build-machine-os-details"
     ./configure \
         --prefix="$GCC_PREFIX" \
         --with-gxx-include-dir="$GCC_PREFIX/include/c++" \
--- a/gcc-4.8/build.sh
+++ b/gcc-4.8/build.sh
@@ -57,5 +57,8 @@ make -j"$CPU_COUNT"
 make install-strip
 rm "$PREFIX"/lib64
 
+#Fix libtool paths
+find -name "$PREFIX"/*.la -print0 | xargs -0  sed -i 's%lib/../lib64%lib%g'
+
 # Link cc to gcc
 (cd "$PREFIX"/bin && ln -s gcc cc)
--- a/gcc-4.8/build.sh
+++ b/gcc-4.8/build.sh
@@ -58,7 +58,8 @@ make install-strip
 rm "$PREFIX"/lib64
 
 #Fix libtool paths
-find "$PREFIX"  -name '*.la' -print0 | xargs -0  sed -i 's%lib/../lib64%lib%g'
+find "$PREFIX" -name '*.la' -print0 | xargs -0  sed -i.backup 's%lib/../lib64%lib%g'
+find "$PREFIX" -name '*la.backup' -print0 | xargs -0  rm -f
 
 # Link cc to gcc
 (cd "$PREFIX"/bin && ln -s gcc cc)
--- a/gcc-4.8/meta.yaml
+++ b/gcc-4.8/meta.yaml
@@ -9,7 +9,7 @@ source:
 
 build:
   detect_binary_files_with_prefix: true # [not linux32]
-  number: 4
+  number: 5
 
 requirements:
   build:
--- a/gcc-4.8/post-link.sh
+++ b/gcc-4.8/post-link.sh
@@ -171,3 +171,4 @@ if [ $SUCCESS -ne 0 ]; then
 fi
 
 cd .. && rm -r "$workdir"
+ln -s ${PREFIX}/lib ${PREFIX}/lib64
--- a/gcc-4.8/post-link.sh
+++ b/gcc-4.8/post-link.sh
@@ -171,4 +171,3 @@ if [ $SUCCESS -ne 0 ]; then
 fi
 
 cd .. && rm -r "$workdir"
-ln -s ${PREFIX}/lib ${PREFIX}/lib64
--- a/gcc-4.8/post-link.sh
+++ b/gcc-4.8/post-link.sh
@@ -7,7 +7,7 @@ if [ "$(uname)" == "Darwin" ]; then
 fi
 
 build_os_md5=( $(md5sum "${PREFIX}/share/conda-gcc-build-machine-os-details") )
-target_os_md5=( $(cat /etc/*-release | md5sum) )
+target_os_md5=( $(lsb_release -a | md5sum) )
 
 # No need to make any portability fixes if
 # we're deploying to the same OS we built with.

### CODE REVIEW

**Comment by ccordoba12** [comment]
@stuarteberg, could you review this one please?

**Comment by stuarteberg** [comment]
Thanks for investigating and diagnosing this, @Zaharid.  I'm trying to think about the consequences of adding `lib64` as a symlink to `lib`.  I have couple of questions.

- Since the symlink is created in `post-link.sh`, shouldn't it be removed in `pre-unlink.sh`?
- What if some other package installs files into `lib64`?  This would cause complications, right? (Granted, I know of no packages that use `lib64` at the moment...)

Your commit message suggests that there might be a "more elegant fix" for #563.  Would that be worth exploring?  If we just patch the `libdir` lines in all the `.la` files, would that suffice?  (Is that ugly?)  I also wonder if there's some option we can pass to gcc's `configure` script that would fix this issue, but I don't see one.

File: gcc-4.8/build.sh [hunk 1]
```diff
@@ -35,7 +35,9 @@ else
     # For reference during post-link.sh, record some
     # details about the OS this binary was produced with.
     mkdir -p "${PREFIX}/share"
-    cat /etc/*-release > "${PREFIX}/share/conda-gcc-build-machine-os-details"
+	# lsb_release can complain about LSB mosules in stderr, so we
+	# ignore that.
+    lsb_release -a 1> "${PREFIX}/share/conda-gcc-build-machine-os-details"
```
stuarteberg: This line looks reasonable, assuming `lsb_release` is always available on all Linux distros.  Is it?  BTW, if you're changing it here in `build.sh`, you should also change it in `post-link.sh`.

groutr: Looks like few linux distributions are LSB certified (mostly enterprise linux distros)
https://www.linuxbase.org/lsb-cert/productdir.php?by_lsb

Zaharid:
Thank you @stuarteberg.

 The motivations for doing the symlink instead of replacing these paths is that I don't understand why they exist in the first place, and it may be something else that depends on them. And I got the idea from the build.sh script that does the same. Indeed, the "more elegant" fix would be removing these paths altogether from wherever they are originated.

The downside is having problems due to the symlink, which would also require that a package installs something in `lib64` which collides with something else in `lib`, which hopefully is rare enough?

Indeed the link should be removed in pre-unlink (even though the typical usage of the gcc package is in a temporary build environment).

gcc-4.8/build.sh
Zaharid: In addition to the distros listed there, I checked that Debian (wheezy) also provides that par command. I'd say that this seems a reasonable solution unless willing to do something much more complex. See http://unix.stackexchange.com/questions/92199/how-can-i-reliably-get-the-operating-systems-name

stuarteberg:
>The motivations for doing the symlink instead of replacing these paths is that I don't understand why they exist in the first place,

Yeah, I wish I knew.  It would be nice to get to the bottom of this.

>and it may be something else that depends on them.

Well, we no that no existing conda packages depend on them, because `lib64` doesn't exist in the current package.  It seems to me that if we get rid of them now, then maybe it will be a non-issue.

>The downside is having problems due to the symlink, which would also require that a package installs something in lib64 which collides with something else in lib, which hopefully is rare enough?

Indeed, that's one potential downside.  I think another downside could be that package maintainers get sloppy with their build script args, and might accidentally use `lib64` instead of `lib`.  If that happens, it will be difficult to undo this decision later if we come up with a better fix.

gcc-4.8/build.sh
stuarteberg: >I'd say that this seems a reasonable solution unless willing to do something much more complex.

Agreed.  Even if few distros are "LSB Certified", it looks like most distros do support the `lsb_release` command.

Zaharid:
Ok, now it replaces the paths in the build script. That's probably cleaner.

Zaharid:
Note that for this to work, everything depending on gcc and libtool needs to be recompiled.

stuarteberg:
>Note that for this to work, everything depending on gcc and libtool needs to be recompiled.

I don't understand.  After all, the `libdir` paths in the current version of the `gcc` package *did not exist*, because `$PREFIX/lib64` doesn't exist.  How could anyone be depending on them in their current form?

Zaharid:
It so happens that some packages can work correctly with the bad paths, but
generate themselves .la files that contain them and break all their
dependencies. And example is this one:

https://github.com/Zaharid/conda-recipes/tree/master/lhapdf

Even if it works with the wrong paths, the .la files that already generated
with the old gcc break its dependencies.

Zaharid:
and if you look for the lib/../lib64 string in the `.la` files present in
an anaconda installation, you see quite a few examples of that.

stuarteberg:
OK, so if I understand you correctly, there may be packages in the wild that need to be recompiled after this PR is merged/released.  But those packages were already broken.  We are doing no harm here, to anyone.

Zaharid:
I think so, yes.

stuarteberg:
OK, in that case this PR looks good to me, but I haven't tested it myself.  That will take some time, and I can't do it today.  I'll try not to let it sit for too long.  Feel free to ping me in a week or two if you haven't heard anything from me.

Zaharid:
Hi,

Is this going to be updated?

gcc-4.8/build.sh
```diff
@@ -55,5 +57,8 @@ make -j"$CPU_COUNT"
 make install-strip
 rm "$PREFIX"/lib64
 
+#Fix libtool paths
+find "$PREFIX"  -name '*.la' -print0 | xargs -0  sed -i 's%lib/../lib64%lib%g'
+
```
stuarteberg: Slight problem here:

On Linux, you can use `sed -i` without specifying a file extension for the backup files, but on OSX, the extension is required. (It's annoying.)

The following will work on both OS X and Linux.

```bash
find "$PREFIX" -name '*.la' -print0 | xargs -0  sed -i.backup 's%lib/../lib64%lib%g'
find "$PREFIX" -name '*la.backup' -print0 | xargs -0  rm -f
```

stuarteberg:
>Is this going to be updated?

Sorry for taking so long to come back to this.  I've now tested your package. It looks good except for the issue I mentioned above regarding `sed -i`.  Once you fix that, this PR is good to go as far as I'm concerned.

FWIW, I built this recipe (plus the `sed -i` fix) and uploaded it to my channel:

```
conda install -c stuarteberg gcc=4.8.5
```

Zaharid:
Please see now

stuarteberg:
LGTM.

@ccordoba12, @msarahan:

I've tested this PR and it looks good.  It's not a huge change, but it sounds like it might be potentially important.  Any chance it can be merged and then added to the `defaults` channel?  I don't see a conda-forge feedstock for gcc.  Should I make one?

msarahan:
I'm happy to merge this here.  I don't know when I'll have time to build it out, though. =(

Feel free to make a feedstock, or to add your builds to an anaconda.org channel.

In general, I am opposed to having a GCC package at all.  I recognize that it makes it much easier for someone to just make things work, but in packaging the compiler, we lose the connection with the system glibc (the package then depends on whatever glibc was on the build machines at the time of build).  We currently have no way to record that information, so linux packages are effectively unknown compatibility.

I would much prefer that we use a docker image for this job (ideally one that we all agree on), and I'd be very excited for any conda package that wraps docker and makes it effectively a black-box compiler.

Zaharid:
Please do ship gcc. It is very useful in my case, where I want to deploy
"modern" software (using C++11) on several different university clusters
each with their own distinct  and old version of gcc (and without much of a
direct access to the administrators of said clusters).

As for libc, old versions should guarantee compatibility with newer ones.
For example I am using a debian 6 docker image to build my packages, and
that seems to work everywhere I care about.

msarahan:
If you use a debian 6 docker image, why do you need our gcc?  http://stackoverflow.com/a/26167100/1170370

Or, just use our docker image?  Look, you're saying "I'm a responsible citizen, so everyone else will be, too!" and that sounds nice, but I just don't buy it.  I also don't really buy into the logic of saying "well, you already have the package, so you should keep maintaining it."  There are better ways to do this, and I have spent at least 80 hours developing our docker image, in the hopes of creating a better way.

Is there a reason why our docker image won't work for you, since you're already using docker?

I'm not opposed to making something that "just works" by listing gcc in a recipe.  I think that's a very nice, explicit expression.  I just don't think that providing gcc as a package is the right way to do it.  If we were shipping libc (not really possible with glibc, as I understand it), this story would be different.

Zaharid:
There is the image I use for building packages that I can build myself, for
which I could indeed use any gcc. But there is some code I just can't build
myself and have to provide a way for people to compile themselves on
"hostile" and diverse environments. For that, having a gcc package is quite
a big help. What version of libc have these users installer is not of
concern because that code only has to run on those particular clusters in
which it is compiled.

msarahan:
I see.  I think I understand what you are asking for: a compiler that you can assume "just works" on an end-user system, without any additional installation or setup steps.

I also understand that libc is not a concern here, because packages built with this compiler won't be redistributed.

Please recognize that that is just one use case - if anyone uploads a package built with this gcc to anaconda.org, then that package's actual compatibility is unknown to anyone but the package creator, until someone downloads it.  Many people are confused when things don't work, and this increases our support burden, because of course people don't ask channel maintainers why packages don't work.

Long-term, I would like to replace the gcc package with a container-based solution.  We may still need the gcc package for places where docker isn't an option for whatever reason (permissions, probably).  It doesn't seem like we'll ever get docker as a conda package, because it will always require sudo access to set up.

PR #572, opened 2016-01-29
fix .binstar.yml for r builds and python 2.7 and 3.4

PeterDSteinberg:
Improvements to .binstar.yml:

 * better support for packages in the r-packages dir
 * python 2.7, 3.4 and 3.5 builds


msarahan:
Thanks!

PR #573, opened 2016-01-30
Update imageio recipe for v1.5

almarklein:
Now that freeimage is available in the main channel (proposed in #335, implemented in #362), the recipe for imageio becomes straightforward.

I verified on Windows, OS X and Linux that imageio picks up the freeimage lib provided by the conda package.

Note that imageio does not depend on a specific version of numpy; is there a way to provide that info in the recipe, to avoid having to build the package against multiple numpy versions?

cc @ccordoba12 @blink1073 @msarahan

### DIFF
--- a/python/imageio/meta.yaml
+++ b/python/imageio/meta.yaml
@@ -1,39 +1,30 @@
 package:
   name: imageio
-  version: 1.2
+  version: 1.5
 
 source:
-  fn: imageio-1.2-linux32.zip  [linux32]
-  url: https://pypi.python.org/packages/source/i/imageio/imageio-1.2-linux32.zip  [linux32]
-  
-  fn: imageio-1.2-linux64.zip  [linux64]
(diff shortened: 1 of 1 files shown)


blink1073:
Sweet!

ccordoba12:
Changes seems straightforward but I don't have an answer to

> Note that imageio does not depend on a specific version of numpy; is there a way to provide that info in the recipe, to avoid having to build the package against multiple numpy versions?

Pinging @ilanschnell about it.

groutr:
It should be fine as is.  If you wanted to build against a specific version of numpy you would have use the the `numpy x.x` feature.  When you build this recipe, does it end up with a dependency on a specific version of numpy, or just a general dependency?

almarklein:
AFAIK it builds against a specific version, but I might be confusing something. To be clear, imageio does *not* require a specific numpy version. (sorry for the late reaction, I've been ill)

PR #582, opened 2016-02-08
Update qutip recipe

nonhermitian:
- Remove -mcf90 flag so that fortran is not required.


ccordoba12:
Why is Fortran not required anymore?

nonhermitian:
It was always an optional install. Since version 3.1 the Python version gives comparable performance, so the fortran files are not so important anymore.

ccordoba12:
Ok, thanks for the explanation.

PR #583, opened 2016-02-09
Correct tensorflow 0.6 recipe

jjhelmus:
Tensorflow version 0.6 is available for Python 2.7 and 3.4, has a single entry point and requires protobuf 3.0.0a3.  This adds support for Python 3.4, corrects the recipe to require protobuf and removes the older non-existent entry point which were available in tensorflow 0.5.  

A recipe for protobuf 3.0.0a3 is also included as this particular version is requires for tensorflow 0.6.

Note that tensorflow is installed from whl packages provided by Google not from source.  There is a shared library in these wheels which may not be compatible with older version of Linux / OS X.

### DIFF
--- a/python/protobuf/meta.yaml
+++ b/python/protobuf/meta.yaml
@@ -1,13 +1,14 @@
 package:
   name: protobuf
-  version: !!str 2.5.0
+  version: !!str 3.0.0a3
 
 source:
-  fn: protobuf-2.5.0.tar.gz
-  url: https://pypi.python.org/packages/source/p/protobuf/protobuf-2.5.0.tar.gz
-  md5: 338813f3629d59e9579fed9035ecd457
+  fn: protobuf-3.0.0a3.tar.gz
+  url: https://pypi.python.org/packages/source/p/protobuf/protobuf-3.0.0a3.tar.gz
+  md5: 6674fa7452ebf066b767075db96a7ee0
 
 build:
+  number: 1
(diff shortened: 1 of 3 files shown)


ahmadia:
LGTM.

msarahan:
If it's OK with you, I'd like to wait until protobuf has a non-alpha (ideally, non-beta) release.

ahmadia:
Ahh, good catch @msarahan.

jjhelmus:
I agree with the desire to use a non-alpha version or protobuf, but the wheel file specifies an exact version of protobuf, 3.0.0a3, as a dependency:

```
Requires-Dist: numpy (>=1.8.2)
Requires-Dist: six (>=1.10.0)
Requires-Dist: protobuf (==3.0.0a3)
Requires-Dist: wheel
```

I take this as an indication that other version are not guaranteed to work.  

From some limited testing protobuf does not actually seem to be required at all, a vendorized copy is included in tensorflow, but if tensorflow is installed using pip as [per the instruction](https://www.tensorflow.org/versions/master/get_started/os_setup.html#pip-installation) a pip install of protobuf 3.0.0a3 is triggered.

msarahan:
Yeah, I agree with your analysis of the situation.  Either they need to fix their dependencies, or they really need that version.  I don't really like the way that Google is doing that, but I appreciate your efforts here.

Merging.

PR #585, opened 2016-02-10
Require Cython for building NumPy and SciPy

jakirkham:
The both have Cython source code that needs to be built and in some cases they may not have the C source code so the safest thing to do is include Cython for building.


msarahan:
Thanks

jakirkham:
Of course.

PR #586, opened 2016-02-10
nomkl: Add recipe for building metapackage.

jakirkham:
Simply adds the `nomkl` package. Seems like some people didn't have this one Windows. ( https://github.com/conda/conda/issues/2032#issuecomment-182322692 ) So, this gives them a way to build it. This, at least, seems to be how it works, but I could be missing something. Feedback welcome.

### DIFF
--- a/nomkl/meta.yaml
+++ b/nomkl/meta.yaml
@@ -1,0 +1,7 @@
+package:
+  name: nomkl
+  version: 1.0
+
+build:
+  track_features:
+    - nomkl


msarahan:
I think this is how it works, but the real reason it doesn't exist is that we don't build non-mkl packages of numpy, scipy and friends.  Those are the packages that get picked up by this feature.  For this package to be useful, there also need to be openblas (or some alternative blas) linked versions, and those need to provide the nomkl feature.

jakirkham:
On Windows, aren't these supplied anyways or are those packages missing too?

msarahan:
The old numpy/scipy builds were repackaged Gohlke packages.  These also had
mkl.  We are pushing towards building everything from source ourselves. At
present, there are no non-mkl builds on windows.

jakirkham:
Ah, ok, I was always curious what BLAS was being used for NumPy and SciPy on Windows. It looked like none was included, but if that were the case performance would suffer. Thanks for explaining what was going on.

Well having an OpenBLAS build could be good. Plus more recent versions of OpenBLAS support CMake for the very reason of having a usable Windows build system. Maybe it would be worth giving that a shot. If I had a working dev environment for Windows, I would give it a shot. At present, I don't. Though it looks like this should be straightforward. ( https://github.com/xianyi/OpenBLAS/blob/develop/appveyor.yml#L38-L39 )

msarahan:
That does look pretty straightforward.  I'll add it to our todo-list.  No promise on when it will happen, though. =(

jakirkham:
Yeah, I know, you guys are swamped.

As a side note, it appears this will use the C kernels on Windows, not the assembly kernels, as the assembly code style is not supported by MSVC. Not sure when or if that will change.

jakirkham:
Probably will still want this though as the `openblas` based recipes now need `nomkl` features added as well and it would be wise to have this here just in case.

PR #588, opened 2016-02-11
libxml2: update to 2.9.3

jschueller:


msarahan:
Thanks.  Can you please explain the gcc inclusion?  Why is it only a build-time dependency? (no libgcc?)

jschueller:
Yeah, I figured all compiled packages should be compiled with the same compiler. Plus I had to build on an older box (Debian Squeeze) to link against an older libc for users of old distros to be able to use my packages. I didn't set the libgcc dependency as it should be a dependency of the zlib recipe, so no need to pull it twice.

msarahan:
Please add the libgcc dependency.  Explicit is better than implicit, in case libgcc is ever removed from zlib.

jschueller:
@msarahan, ok

msarahan:
Thanks!

PR #589, opened 2016-02-11
Add otsubset package

jschueller:


msarahan:
Since this package has CMake, can you modify this recipe to build from source on Windows, instead of repackaging binaries?

jschueller:
@msarahan, see my remark for OpenTURNS, which is difficult to build on windows, here it should be compiled in the same manner otherwise it results in an mingw/msvc c++ ABI incompatibility.

PR #590, opened 2016-02-11
Openturns

jschueller:


msarahan:
Since this package has CMake, can you modify this recipe to build from source on Windows, instead of repackaging binaries?

jschueller:
@msarahan we use to cross-compile it from Linux/MinGW-w64 as building with msvc is so long to set up for the different 32/64 variants. Also supporting Python 3.5 requires the vc2015 runtime for which the compilation has not been investigated yet (succeeding to build on vc2010 was already a nightmare).

msarahan:
Wow, hats off to you for cross-compiling.  Getting that set up always scared me.  I would really like to have actual build code in recipes, though, rather than repackaging instructions.  If you don't have access to a Windows build machine, I can develop this and test it.  CMake generally makes it extremely easy.  There are a few recipe snippets that might be helpful here:  https://github.com/conda/conda-build/wiki/Windows-recipe-patterns

If you need me to do it, I can't say exactly when it will happen.  I am pretty busy these days, but I'll try to work it in.

jschueller:
@msarahan Again, we haven't successfully built Python bindings with msvc, we only provide the c++ library for msvc10. There's the need to set up all the dependencies with the same compiler: libxml2, muparser, nlopt, boost, tbb ; which all are not so easy to compile with msvc, we had to rewrite custom cmake lists for some.

Please don't do that, it's really long and painful.

msarahan:
Ok. I'll merge this. I still think it is better to build, though. Fwiw, I have taken a similar strategy of custom cmake files when building serf (dependency of svn) and ICU.  If like to develop those scripts and submit them upstream, but for now they are rather windows specific.  If it's OK with you, I might copy stuff from the file you linked to to improve the other conda recipes here.

jschueller:
@msarahan, you're welcome to do so

PR #594, opened 2016-02-16
Refine GLPK recipe

Juanlu001:
See https://github.com/conda/conda-recipes/pull/481 for previous discussion. The code is a merge between @aebrahim and @snorfalorpagus recipes. Notably, adds some testing and fixes the binary paths.

### DIFF
--- a/glpk/bld.bat
+++ b/glpk/bld.bat
@@ -1,14 +1,7 @@
-cd w%ARCH%
-copy config_VC config.h
-
-nmake /f Makefile_VC
-
-mkdir %PREFIX%\bin
-mkdir %PREFIX%\include
-mkdir %PREFIX%\libs
-
-copy glpsol.exe %PREFIX%\bin
-copy ..\src\glpk.h %PREFIX%\include
-copy glpk.lib %PREFIX%\libs
+copy src\glpk.h %LIBRARY_INC%\glpk.h
+copy "w%ARCH%\\*.lib" %LIBRARY_LIB%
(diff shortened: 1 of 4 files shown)


aebrahim:
I think using winglpk as done here might be problematic because it uses pre-compiled glpk instead of using the appropriate VS version to compile it during the build step.

This would make it impossible to link glpk with python libraries in python 3.5, which uses VS 2015 which changed the c runtime in an incompatible way.

Juanlu001:
You are right, it was fine for my own use case but using binaries does not make sense here. Just made the edit.

glpk/bld.bat
```diff
@@ -1,14 +1,7 @@
-cd w%ARCH%
-copy config_VC config.h
-
-nmake /f Makefile_VC
```
msarahan: I thought the plan was to keep the compiled version, but add the tests?

Juanlu001: Sorry, I should have restored this too when rolling back the first commit. Committing.

aebrahim:
It looks like there are still some issues:
 - glpk_4_57.lib doesn't exist after compiling
 - there will be no dll files after compiling

 To resolve them I've made #595.

I couldn't figure out a good way to test compiling and linking against glpk on windows. If I do I'll put it in a separate PR.

aebrahim:
OK, in #595 I made a dummy cython package which links against the library. This will also ensure that the package works with the various python versions. That should be a roughly equivalent test to the version.c compiling here while also being cross platform.

Juanlu001:
Closing this in favor of #595

PR #595, opened 2016-02-17
GLPK: fix windows paths and add glpsol test

aebrahim:
This fixes some bugs in #594


glpk/meta.yaml
```diff
@@ -17,6 +17,16 @@ requirements:
   build:
     - gmp      [not win]
 
+test:
```
msarahan: Need to add Cython as a test-time dependency.  

This is a nice approach to hide compilation complexity, by the way.  A good example for other projects that have this sort of thing to test.

msarahan:
@Juanlu001, are you OK with this PR?  If so, would you rather roll this into your PR (@aebrahim can submit a PR against your branch that your PR is created from), or just merge this?  I like his cython approach since it will facilitate testing on Windows.  The alternative would be to figure out the call to compile things on windows (cl.exe instead of cc, not exactly sure how much more is required.)

Juanlu001:
@msarahan Yes, totally! Whatever works best. @aebrahim has already done the work so we can close mine and merge this.

glpk/meta.yaml
aebrahim: Thanks I'm glad you like it!

I tried adding the following block:
```
test:
  requires:
    - cython
```

Any idea why I'm getting this error message?
```
BUILD END: glpk-4.57-vc14_0
TEST START: glpk-4.57-vc14_0
Solving package specifications: ..
Error: Unsatisfiable package specifications.
Generating hint:
[      COMPLETE      ]|##################################################| 100%

Hint: 'glpk' has unsatisfiable dependencies (see 'conda info glpk')
```

msarahan: Ugh, we have some issues with test-only dependencies.  If you don't mind, please add it as a runtime dependency also, so that it gets installed correctly.  I don't want our test-dependency issue to hold up your PR.  Add a comment about this to the recipe, if you'd like.

Ping @kalefranz @mcg1969 - if either of you run into code that might address this problem, let me know.

aebrahim: I get the same error with cython as a runtime dependency as well,
unfortunately.

I've worked around the issue by pre-generating the C file and including it.

msarahan:
Great, thanks. Merging.

We will look into the unresolvable dep issue.  I could not reproduce locally with

Conda create -n cythontest python=3.5 cython

Does that command work for you?  If not, you might need to update conda.

aebrahim:
I can install cython just fine on its own. The issue occurs when I run
```conda build glpk``` from the conda-recipes directory

Perhaps it's a bug with the VS14 feature not being satisfied correctly?

In any case, thanks for merging!

aebrahim:
On an unrelated note, if you don't mind my asking, is there an easy way to
get binaries of these recipes for all 3 platforms?

I ask because I maintain a python package (cobra) which needs to be linked
to glpk (as in it's not a pure python package). I personally don't use
anaconda, but I have gotten many bug reports users who do and are unable to
install it, especially on windows (by contrast, for regular python I have
appveyor build wheels which are uploaded to pypi). But even after getting
the recipe working (including for the glpk dependency), I'm unclear on how
the package should be deployed.

PR #596, opened 2016-02-17
perl, scala, java root-level directories

kalefranz:
also moved python interpreters to python directory


kalefranz:
If there are no objections in the next 24 hours or so, I'm going to go ahead and merge this.

msarahan:
+1

PR #597, opened 2016-02-17
create c-compiled root directory

kalefranz:
With the initial commit, this PR includes all packages that have `gcc` in build requirements or set `CC` in `build.sh`.  Many, many more packages need to move to this `c-compiled` directory.

Also, completely open to names other than `c-compiled`.  Does anybody have anything better?


msarahan:
maybe ```shared-libs```

that doesn't really include standalone exe's though.

kalefranz:
@ilanschnell what are your thoughts for the name of this directory?

kalefranz:
@msarahan 

> that doesn't really include standalone exe's though.

Also, some things are shared-libs, some things are standalone, and some things are both :-/

kalefranz:
Let's play Jeopardy with this.

Definition: Packages that--in their distributed form--are primarily compiled by a c/c++ compiler to a binary form that does not rely on an interpreter or virtual machine.

Now what is the word?

kalefranz:
How about `c-cxx`?

kalefranz:
How about just `c`?

kalefranz:
I'm leaning toward `c` being the most logical. I think users of this repo are naturally able to connect with the language applications and libraries are written in. This would be for "pure" `c` and `c++` libraries and apps. Packages that contain compiled c/c++ code but are intended for use by other languages via foreign function interfaces belong in the "user language" namespace.  E.g. numpy belongs in the python directory.

I'm lumping `c` and `c++` together here just into `c`. At this point, I don't see a need to disambiguate the two.

Also, interpreters and compilers, regardless of the language they're written in, belong in their namesake's directory.  E.g. CPython and pypy both belong in the python directory; both go 1.4 and 1.5 belong in a (future) go directory.

PR #598, opened 2016-02-17
h5py: Update to 2.6.0.

jakirkham:


msarahan:
Great to see Windows compiled from source.  Are there any changes to bld.bat necessary?

jakirkham:
Good call, fixed the `bld.bat` to use `setup.py`. Do I need some selectors for python on Windows?

Have tested on Mac and Linux. Unfortunately, I don't have Windows to test that case. Though we do provide `hdf5` for Windows now. So *should* work.

msarahan:
Nope, no selectors necessary for python.  It is sufficient to have it listed as a runtime dependency.

You might want to move unittest2 down to a test-time requirement, but that's not terribly important.  I'll go ahead and merge.

jakirkham:
Oh, ok. I was just about to grab [these]( https://github.com/conda/conda-recipes/blob/master/boost/meta.yaml#L10-L14 ). If we don't need them, then nvm. Otherwise, I can put them in a new PR.

Yeah, you're probably right with unitest2. It was already there so I left it. I think `h5py` lets one run the tests from Python (like NumPy) so I think that is why it was added as a runtime dependency previously.

msarahan:
Those are necessary when you have a library that doesn't have Python as a runtime dependency to differentiate the compiler version.  So HDF5 should have those - same for NetCDF4.  However, h5py doesn't need them because there's a different version for each Python version, by virtue of the runtime python dependency.

jakirkham:
Right, I understand now. Thanks for clarifying.

PR #599, opened 2016-02-17
h5py: Run test suite during testing

jakirkham:
Takes ~1s to run.


msarahan:
Thanks!

jakirkham:
Yep, I was checking to see if we were already doing this with `run_test.py` and it wasn't the case so figured it would be a simple addition.

PR #600, opened 2016-02-18
Tensorflow version 0.7.0 recipe

jjhelmus:


jjhelmus:
Note that there are issues with tensorflow 0.7.0 on linux-64 with Python 3.5 tensorflow/tensorflow#1143

jjhelmus:
Not sure if this is worth merging, there are some issues with the 0.7.0 tensorflow whl files provided by Google which will cause the package not to work in a number of cases.  Building the package from source using bazel is likely the best option.

msarahan:
Cool - looks like you're on top of debugging this in that thread.  Let us know when the 0.7.1 release is out, and we'll merge that in.

Eventually I agree I'd like to build from source.  Having a working copy in the immediate future is more important, though.

jjhelmus:
Updated this with version 0.7.1 wheels.  Need to update the protobuf recipe to 3.0.0b2 which will be coming shortly.

jjhelmus:
Also should say a recipe using bazel to build from source in planned, just need a bit more time to work out the details.

msarahan:
Awesome!  I'm following your work over at conda-forge.  If it's OK, we'll guinea-pig this project for moving towards a different way of doing recipes: host them primarily on conda-forge, but have submodules here.

jjhelmus:
Sounds good to me and also less work since conda-forge already has a protobuf 3.0.0b3 recipe.

PR #602, opened 2016-02-18
add configure step to git recipe

msarahan:
Without the configure step, the build process does not seem to respect the LIBRARY_PATH variable, and libraries end up not getting found.  I am guessing that the configure step records LIBRARY_PATH into its build steps.

This may have been masked in earlier builds by system libraries fulfilling needs (libssl is the one that kills the build for me).  This problem was highlighted when trying to build with a minimal docker image (https://hub.docker.com/r/msarahan/conda_builder_linux/)

Ping @frol @ccordoba12 - I know some people don't like the configure step because it introduces dependencies (m4? autotools?).  Is there a better way to do this that you know of?

### DIFF
--- a/git/build.sh
+++ b/git/build.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-export C_INCLUDE_PATH="$PREFIX/include"
-export LIBRARY_PATH="$PREFIX/lib"
+export C_INCLUDE_PATH="${PREFIX}/include"
+export LIBRARY_PATH="${PREFIX}/lib"
 
 # NO_TCLTK disables git-gui
 # NO_PERL disables all perl-based utils:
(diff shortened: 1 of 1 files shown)


msarahan:
PS, see interesting comment at http://stackoverflow.com/a/4293671/1170370:

```
Note: when linking libraries, ld by itself does not look for libraries in either LIBRARY_PATH or LD_LIBRARY_PATH. It's only when gcc invokes ld that LIBRARY_PATH becomes used. (Learned this the hard way.) – Rufflewind Jul 2 '14 at 0:34 
```

frol:
@msarahan Looks good to me if it compiles on your end.

ccordoba12:
+1 from me too. There are already recipes here for `m4` and `autotools`, so maybe you should make git depend on those for its `build` step.

PR #609, opened 2016-02-22
fix libffi build.sh shellcheck warnings

jhoblitt:


jhoblitt:
I have added an additional change set to fix a problem with the header include path that I discovered this morning.  I believe the presence of system headers has been masking this issue.

PR #610, opened 2016-02-22
update pytest-cov

groutr:


msarahan:
Nice simplification! You sure it still works?

groutr:
It was regenerated by conda skeleton.  It builds.
I'm using it at the moment to run conda's test suite and it appears to work.

msarahan:
All right. Thanks!

PR #616, opened 2016-02-27
Make Qt4 work side by side with Qt5

ccordoba12:
This is necessary to make these Qt versions work together without problems.

This PR works like this:

1. Install shared data and binaries under `$PREFIX/lib/qt4`
2. Make symlinks for all binaries installed into `$PREFIX/lib/qt4/bin` to `$PREFIX/bin`, with the `-qt4` suffix added to avoid clashes with Qt5.
3. Add a bash script to call `qmake-qt4` with its full path because if we create a symlink for it, `qmake-qt4` can't find `qt.conf`.

I'm still testing if this works on all situations, so please don't merge it. I'll merge it when it's ready.

Pinging @msarahan about this one.


msarahan:
This looks OK, but we will need to have patches or hacks in many build scripts to adjust for the qmake-qt4 binary.  Perhaps you could instead just add that folder to PATH inside that build script?  Regardless, it stinks that we have to change recipes elsewhere for this.  I don't see a better way, though.

How much do headers need to be sectioned off into separate folders?

On Windows, this is a tiny bit easier, because the DLLs have the major version number in their filename.

ccordoba12:
> Perhaps you could instead just add that folder to PATH inside that build script?

That's another option. For PyQt4 I'm calling `qmake-qt4` directly because it has an option for that. What other recipes do we need to modify?

> On Windows, this is a tiny bit easier, because the DLLs have the major version number in their filename.

The problem is not with libraries (Linux `so` libs are versioned too). It's with binaries, and shared data (mkspecs, plugins, etc), which have the same names in Qt4 and Qt5.

mingwandroid:
Your solution is the same one that ArchLinux went for (except they did it the other way round, for historical reasons, with qt5 being the one that got the symlinks):

https://projects.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/qt5

.. so there's some precedent for this approach.

ccordoba12:
@msarahan, that would be another option: create symlinks for Qt5 binaries only, and leave the same names for Qt4. What do you think?

ccordoba12:
Sorry, I meant "create special names for Qt5 binaries only"

msarahan:
Meh, I don't have strong opinions here.  I'm fine to follow Arch's example, though I'd prefer some less cutting-edge linux to follow, for the sake of being more conformist.  Is there an established way that Red Hat or Debian/Ubuntu handle this?

ccordoba12:
Ubuntu also adds suffixes to distinguish Qt binaries, and they are called `qmake-qt4`, etc.

ccordoba12:
I think this is ready. I tested new packages created with these recipes in Linux and Mac, and they are working as expected.

I also changed the only recipe that was using `qmake` directly (shiboken).

Thoughts @msarahan?

ccordoba12:
Ok, since I haven't heard any objections, I'm merging this one :-)

msarahan:
Sorry, I'm swamped.  You're fine here.  Thanks for merging.

ccordoba12:
Not a problem :-)

PR #624, opened 2016-03-05
Qscintilla2

dschreij:
Working recipes for MacOS and Windows (32/64 bit) to build qscintilla2 against qt4/pyqt4.
Once (py)qt5 is released I will try build it against that too.

### DIFF
--- a/qscintilla2/bld.bat
+++ b/qscintilla2/bld.bat
@@ -1,23 +1,63 @@
+@echo off
+:: MSVC library for Python 2.7 is not found automaticaly if it is 
+:: installed in the user's AppData folder. Add it to PATH
+setlocal EnableDelayedExpansion
+
+if "%PY_VER%"=="2.7" (
+	set VC9DIR=%LOCALAPPDATA%\Programs\Common\Microsoft\Visual C++ for Python\9.0
+	IF EXIST !VC9DIR! ( 
+		set PATH=!VC9DIR!;"%PATH%"
+		if %ARCH% == 32 call vcvarsall.bat x86
+		if %ARCH% == 64 call vcvarsall.bat amd64
+		:: vcvarsall does something strange with PATH, preventing xcopy to be found later. Fix this
+		set PATH=!PATH!;%SYSTEMROOT%\system32
+	)
+)
+if errorlevel 1 exit 1
+
 :: Set QMAKESPEC to the appropriate MSVC compiler
 set QMAKESPEC=%LIBRARY_PREFIX%\mkspecs\win32-msvc-default
 
+@echo ====================================================
+@echo Building Qscintilla2
+@echo ====================================================
 :: Go to the source folder and enter the Qt4Qt5 dir
 cd %SRC_DIR%\Qt4Qt5
 :: Use qmake to generate a make file
 %LIBRARY_BIN%\qmake qscintilla.pro
+if errorlevel 1 exit 1
+
+@echo Compiling
 :: Build and install
 nmake
+if errorlevel 1 exit 1
+@echo Installing (copying)
+@echo ====================================================
+@echo PATH = %PATH%
+@echo ====================================================
 nmake install
+if errorlevel 1 exit 1
 
+@echo ====================================================
+@echo Building Python bindings
+@echo ====================================================
 :: Python bindings
 :: Go into the Python folder
 cd %SRC_DIR%\Python
 :: Use configure.py to generate a MAKEFILE
+@echo Configuring with python
 %PYTHON% configure.py
+if errorlevel 1 exit 1
 :: Build and install
+@echo Compiling python modules
 nmake
+if errorlevel 1 exit 1
+@echo Installing python modules
 nmake install
+if errorlevel 1 exit 1
 :: The qscintilla2.dll ends up in Anaconda's lib dir, where Python
 :: can't find it for import. Copy it to the bin dir
 :: (as indicated at http://pyqt.sourceforge.net/Docs/QScintilla2/)
 copy %LIBRARY_LIB%\qscintilla2.dll %LIBRARY_BIN%
+if errorlevel 1 exit 1
+@echo finished
--- a/qscintilla2/bld.bat
+++ b/qscintilla2/bld.bat
@@ -1,0 +1,3 @@
+set QMAKESPEC=%LIBRARY_PREFIX%\mkspecs\win32-msvc-default
+
+%LIBRARY_BIN%\qmake qscintilla.pro
--- a/qscintilla2/bld.bat
+++ b/qscintilla2/bld.bat
@@ -15,8 +15,18 @@ if "%PY_VER%"=="2.7" (
 )
 if errorlevel 1 exit 1
 
+:: Determine Qt Major verion (4 or 5)
+for /f "tokens=4" %%i in ('%LIBRARY_BIN%\qmake -v ^| find "Qt version"') do (
+	set QT_MAJOR_VER=%%i
+	set QT_MAJOR_VER=!QT_MAJOR_VER:~0,1!
+)
+if [!QT_MAJOR_VER!] == [] Exit /B 1
+
+@echo Building for Qt!QT_MAJOR_VER!
+
 :: Set QMAKESPEC to the appropriate MSVC compiler
-set QMAKESPEC=%LIBRARY_PREFIX%\mkspecs\win32-msvc-default
+:: Not really necessary in windows
+:: set QMAKESPEC=%LIBRARY_PREFIX%\mkspecs\win32-msvc-default
 
 @echo ====================================================
 @echo Building Qscintilla2
@@ -46,7 +56,7 @@ if errorlevel 1 exit 1
 cd %SRC_DIR%\Python
 :: Use configure.py to generate a MAKEFILE
 @echo Configuring with python
-%PYTHON% configure.py --pyqt=PyQtn
+%PYTHON% configure.py --pyqt=PyQt!QT_MAJOR_VER!
 if errorlevel 1 exit 1
 :: Build and install
 @echo Compiling python modules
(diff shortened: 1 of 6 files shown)


ccordoba12:
@dschreij, thanks a lot for working on this! i have a couple of comments, but I'll do them in your recipes.

qscintilla2/build.sh
```diff
@@ -0,0 +1,30 @@
...
+	make
+	# and install it
+	make install
+
+	## Build Python module ##
+
+	# Go to python folder
+	cd ${SRC_DIR}/Python
+	# Configure compilation of Python Qsci module
+	${PYTHON} configure.py --qmake=${BIN}/qmake --sip=${BIN}/sip --qsci-incdir=${PREFIX}/include --qsci-libdir=${PREFIX}/lib --spec=macx-llvm --no-qsci-api
```
ccordoba12: `qmake` binary is going to be renamed to `qmake-qt4` (see pull request #616 for the rationale).

New Qt4 packages with this change are going to be released shortly (hopefully by the end of the week).

qscintilla2/bld.bat
```diff
@@ -0,0 +1,63 @@
...
+:: Set QMAKESPEC to the appropriate MSVC compiler
+set QMAKESPEC=%LIBRARY_PREFIX%\mkspecs\win32-msvc-default
+
+@echo ====================================================
+@echo Building Qscintilla2
+@echo ====================================================
+:: Go to the source folder and enter the Qt4Qt5 dir
+cd %SRC_DIR%\Qt4Qt5
+:: Use qmake to generate a make file
+%LIBRARY_BIN%\qmake qscintilla.pro
```
ccordoba12: Also `qmake` binaries will be renamed eventually, but this change could take more time than the OSX one :-)

ccordoba12:
@dschreij, question: what about building this on Linux? Could you work on that too?

dschreij:
Yes, I definitely could an was actually planning to. I think Linux will be the easiest of the bunch judging from past experiences. I don't have a Linux VM working at the moment due to lack of disk space, but I wanted to set one up somewhere this week.
Any specific demands concerning this? I was thinking of going for Ubuntu 15.10 or give elementary.io a spin, but I see a lot of recipes that have special additions for CentOS for instance.

qscintilla2/build.sh
dschreij: Alright, let me know what I can do from here. I also successfully managed to compile Qscintilla2 against (Py)Qt5 with a minimum of changes to this script, but this was for Homebrew Python/Qt. It shouldn't be that much of a stretch for the Anaconda environment. Good to hear Anaconda is also pushing towards Qt5! Once it does so officially, I can see if I can adapt these build scripts accordingly

ccordoba12:
We (at Continuum) use CentOS 5 and gcc 4.4 to compile our packages, but the recipes added here doesn't need to adhere to those specs.

dschreij:
Ok, I have changed the build.sh script in such a way that it also works on Linux. The required changes were only minor. I uploaded the builds to anaconda.org as well for now.

msarahan:
Thanks!  It is not clear to me if this recipe belongs in the python folder (because it provides an add-on to PyQt), or whether it belongs in the root (for now), to be moved to some kind of "compiled C/C++ system library) folder.  @ccordoba12 @kalefranz do you have a quick opinion?

ccordoba12:
This should probably go into the Python directory, but please don't merge until our new Qt4 packages for Linux and Mac are out, so that @dschreij can change his recipes to adjust to them.

ceholden:
Hi everyone -- could we merge this recipe or move it over to conda-forge? I'm hoping to target it as a dependency

I can confirm that the Linux build is working on the most recent QScintilla version (2.9.3) once I removed the manual specification of lib/include, but I haven't been able to test the Windows / Mac side of things (something the conda-forge CIs will be able to do). I'd be happy to do the work or assist in getting it over to conda-forge if that helps things along.

Thanks @dschreij for the excellent recipe!

dschreij:
Hi guys, thanks for the kind words ceholden. I've been waiting for your signal to also convert this recipe for usage with qt5 but thus far I haven't been able to get a hold of any (official) qt5 packages. Please let me know once qt5 is released in the wild on anaconda and I can see if I can make it work with both qt4 and 5. I recently also tried to update the QScintilla package on our conda channel to 2.9.3, but I received the message that the sip version which was on conda then was too old, although this was a while ago. I'm glad @ceholden managed to get it to work.

ccordoba12:
@dschreij, I think the best thing to do is to move this recipe to conda-forge:

https://github.com/conda-forge/staged-recipes/

Qt5 packages are available for testing in our qttesting channel:

https://anaconda.org/qttesting/qt/

dschreij:
Alright @ccordoba12! I built QScintilla2 with (Py)Qt 5.6 from your testing channel, and this went without problems. QScintilla2 is running great, so I'm very happy about that. I only built it for Mac for now, but I expect the process under Windows and Linux to be little different. 

I had to change a couple of things to the recipe though, which I'd like to ask a few questions about before I move the recipe to conda-forge:

- The Qt header files appear to have moved from `${PREFIX}/include` to `${PREFIX}/include/qt` (thus in an extra subfolder qt inside `include`). Has this changed for all qt versions, or only the newer qt5? If this is different for 4 and 5, shall I place the original recipe in a separate folder `qscintilla2-qt4` and the new one in `qscintilla2` (as you've proposed before) in conda-forge?

- I had to pass the extra flag `--pyqt=PyQt5` to the `python configure.py` command, because this script defaults to PyQt4. What is the best way to handle this? It seems a bad way to hardcode this flag into the build script. Is there a variable in the conda build environment from which you can determine which Qt4 version is currently used, and that we can do something like `--pyqt=PyQt${QT_MAJOR_VER}`

That's it. Once I know about the above I am ready to move verything over to conda-forge.

dschreij:
Alright, I have solved the second problem by extracting the PyQt version string from the qmake that is present in the environment, with qmake -v

~~~bash
QT_MAJOR_VER=`qmake -v | sed -n 's/.*Qt version \([0-9])*\).*/\1/p'`
if [ -z "$QT_MAJOR_VER" ]; then
	echo "Could not determine Qt version of string provided by qmake:"
	echo `qmake -v`
	echo "Aborting..."
	exit 1
else
	echo "Building Qscintilla for Qt${QT_MAJOR_VER}"
fi
~~~
Then you can use this value to compose the configure.py string.
~~~bash
${PYTHON} configure.py --pyqt=PyQt${QT_MAJOR_VER}  #etc.
~~~
The script should thus now be qt4 and qt5 safe, although of course the specifics of versions used need to be indicated in the meta.yaml file. If version constraints are indicated there, the build.sh should do the rest and work regardless.

Regarding the first problem. I just assume that /include/qt is the definitive new location where the qt header files are placed, so I'll just push this script as is to  conda-forge?

ceholden:
@dschreij there are a few stylistic standards that need to be adhered to for conda-forge recipes, including the use of Jinja templating for version information and the addition of a list of recipe maintainers to the `meta.yaml`. The `conda-forge/staged-recipes` repo integration services will check recipes added during a PR for these constraints and give you feedback on what needs changing so you can iterate on the recipe to get it compliant with the standards.

There is a template or example recipe file in the `staged-recipes` repo with more info: https://github.com/conda-forge/staged-recipes/blob/master/recipes/example/meta.yaml

It looks like @ccordoba12 is at work adding Qt5 and PyQt5 to `conda-forge` -- by [adding a Qt5 build to the existing](https://github.com/conda-forge/staged-recipes/pull/606) `conda-forge/qt-feedstock` and by [adding a new recipe for PyQt5](https://github.com/conda-forge/staged-recipes/pull/727). Per [conda-forge FAQ question 7](https://github.com/conda-forge/staged-recipes/wiki/Frequently-asked-questions), they really want all dependencies to come from other `conda-forge` channels, so I think we'll need to wait until the Qt5 situation in `conda-forge` is settled.

As for the include location, `include/qt` is the spot for the conda-forge's existing qt4 build so I think you have it right.

dschreij:
Alright, have tried to adapt the meta.yaml format of the conda-forge example you linked as best as I could, but there were some parts that confused me and at which I couldn't really follow the format:

- I have separate sources for Windows (a .zip) and for Linux/OSX (.tar.gz) which each have their own MD5 (sadly sourceforge only provided SHA1 or MD5, not the preferred SHA256). How do incorporate that into the jinja template strings?
- The download filenames are not formatted as {{name}}-{{version}}, so I left them in literally.
- At the build section, I left out the script entry as recommended when the build process is too complex.
- At recipe maintainers, should I enter my anaconda.org or Github username (dschreij), or my first and last name CamelCased (DanielSchreij)

I you think this recipe is good enough to be placed at conda forge, or well on its way, let me know and I will integrate the changes there and send a PR.

ceholden:
This all looks great to my eye as it is, and good enough to let the linter catch anything else. It might be picky with the spacing of the platform flags and version pins, but I don't know off hand.

I'd keep with the zip and tar.gz since the other Qt recipes also split the sources, though I'm not sure if you couldn't get away with just the tar.gz. The package file name is somewhat unfortunate, but it might be nice to add in just the `{{ version }}` for the future so you only have to change it in one place.

Your Github username is the right thing to add for the maintainer slot. When the recipe builds correctly and passes the CI services on `conda-forge/staged-recipes`, some magic happens behind the scenes to create a `conda-forge/qscintilla-feedstock` repo which you will be added as administrator over so it will need your Github username.

I'd go ahead and submit it over at conda-forge and we can continue with any changes over there. Thanks again for your work -- getting this in place will be one more dependency solved that should enable me to build and distribute a program through conda that's been giving me headaches compiling from scratch on different machines!

dschreij:
Ok, it's now up and running at conda-forge/stage-recipes and the linter already has found some flaw, but I will have to look at this later. Let's close this discussion here, and continue it over at https://github.com/conda-forge/staged-recipes/issues/1395 if necessary.

ccordoba12:
:+1:

PR #628, opened 2016-03-08
Removing Misuse of Build Versions

tinyheero:
This pull request removes the OS specific build versions which isn't what these values are meant for.

### DIFF
--- a/r-packages/r-magrittr/meta.yaml
+++ b/r-packages/r-magrittr/meta.yaml
@@ -9,6 +9,8 @@ source:
   url:
     - http://cran.r-project.org/src/contrib/magrittr_1.5.tar.gz
     - http://cran.r-project.org/src/contrib/Archive/magrittr/magrittr_1.5.tar.gz
+
+
   # You can add a hash for the file here, like md5 or sha1
   # md5: 49448ba4863157652311cc5ea4fea3ea
   # sha1: 3bcfbee008276084cbb37a2b453963c61176a322
@@ -19,8 +21,7 @@ source:
 build:
   # If this is a new build for the same version, increment the build
   # number. If you do not include this key, it defaults to 0.
-  number: 2 # [osx]
-  number: 1 # [not osx]
+  # number: 1
 
   # This is required to make R link correctly on Linux.
   rpaths:


groutr:
Can you explain your line of thinking?

msarahan:
These are incremented by convention when something about the recipe changes.  The changes may not affect all platforms.  It is thus common to see build numbers that differ between platforms.  Whether the build number should just simply be bumped for all platforms at once is up for debate.  However, that implies forcing a new build on every platform, so we prefer to keep it to only the one(s) that have changed.

tinyheero:
Hi.

I don't understand why is there a need for a OS specific build version. The recipe hasn't changed and the versions of the software is exactly the same. I was under the impression the build number is used when the recipe has changed, but the version of the software is exactly the same. In this particular case, these recipes should be created from CRAN skeletons:

```
conda skeleton cran magrittr
```

So having a build number even based on a skeleton doesn't make sense to me?

This also adds problems when trying to bring the environment to a different OS. For instance, I built an environment in Linux 64, and then I brought it to my Mac OS X. Even though for instance, the binaries for r-magrittr may exist for both Mac OS X and Linux OS, it will complain that there exists no `r-magrittr-1.5-r3.2.2_1.tar.bz2` for Mac OS X because the build number is 1 for non-OSX. Yet, there exists a `r-magrittr-1.5-r3.2.2_2.tar.bz2` for Mac OS X which is the exact same r-magrittr version yet it gets a build number of 2 for being a Mac OS X? So the only thing that is different between this is the fact that there is a different build number? 

Additionally, only a subset of r-package recipes seem to have this behaviour? I also do not see any of the bioconda r-recipes having this behaviour?

groutr:
The purpose of the build number is exactly that, to differentiate different builds of the same version of a piece of software.  Usually, the build number is bumped if something in the build script changes.  In this case, it looks like the last change to the build script removed an OSX specific environment variable.

How did you "bring your environment" from linux-64 to osx?

tinyheero:
But it isn't a "different build" of the same version. It's the exact same build recipe. Shouldn't the build number be tied to some recipe change for the same version of a piece of software? If I build the software on OS X and I use the **same recipe** to build the software on Linux 64, why should the build number be different? 

What I did was:

1. On a linux 64 machine, I create an conda virtual environment installing R packages for a particular project:

    ```
    conda create --name test r
    source activate test
    conda install r-magrittr
    ```

This installs `r-magrittr: 1.5-r3.2.2_1`. Note the build number is 1. 

1. Export my environment:

    ```
    conda env export > environment.yml
    grep magrittr environment.yml
    ```

1. Bring this environment.yml file to my Mac OS X. Try to create the environment:

    ```
    conda env create environment.yml
    ```

Among the errors will be:

    Error:  Packages missing in current osx-64 channels:
    .
    .
    .
    r-magrittr 1.5 r3.2.2_1

Because the Mac OS X version is r-magrittr 1.5 r3.2.2_2. 

I realized that I likely won't be able to fully reproduce an environment going between Linux 64 and OS X since certain libraries will be OS X specific (e.g. libxml2, cairo). But I was thinking I could just maintain the R package versions between different OS...

msarahan:
Yes, great point.  This is a shortcoming in the export method.  It records the exact version including build number, which is not OS portable, as you have found.  The solution is to edit your environment.yml and remove overly specific things, like build number.

msarahan:
Issue moved to [conda/conda-env #230](https://github.com/conda/conda-env/issues/230) via [**ZenHub.io**](https://www.zenhub.io/)

tinyheero:
I don't think this should be closed. Having the export record build number is good and makes sense.

What doesn't make sense is having a build number tied to a platform like the r-magrittr recipe. Again, recipe hasn't changed. Just somewhere along the along, there was a build number added for OS X and a build number of Linux 64 in the recipe. Yet the rest of the recipe remains the same. There isn't a need to have a build number indicate a whether the binary is OS X or Linux 64. Conda should take care of figuring out whether binary is OS X or Linux 64.

msarahan:
This isn't "closed" - it is moved to conda-env.

Conda does take care of figuring out which binary to install, but it can't figure out what differences between build numbers are, especially across platforms.  To say something like:

    I have build 2 on mac, map that to the same thing on Linux

makes no sense.  They are free to change independently, and the build number carries absolutely no information across platforms.

tinyheero:
Sorry I don't understand...if the build number carries no information across platforms, then why is the build number assigned to be platform specific?

Snippet from r-magrittr recipe:

```
build:
  # If this is a new build for the same version, increment the build
  # number. If you do not include this key, it defaults to 0.
  number: 2 # [osx]
  number: 1 # [not osx]
```

Same recipe is being used to build on both Linux 64 and Mac OS X. Yet when we build on Linux 64, it gets the build number of 1. And on Mac OS 64, using the same recipe, it gets the build number of 2. There is nothing OS X specific about recipe for build number 2.

msarahan:
There may not be any difference in the builds at this current point in time.  At some point in time in the past, they may have differed.  The only ways to "synchronize" them and have them with with same build number after that point are either:

   - update the recipe to a new version of software, at which point it is best practice to reset the build number to 0
  - arbitrarily skip build numbers so that they're uniform again.

The build number is a reflection of history, not indicative of current state.

tinyheero:
Ok fair enough. I guess we'll have to wait until a new version of r-magrittr comes out before we can sync the build numbers.

tinyheero:
One solution I just thought about was incrementing the build number to be 3 and make it platform independent. This would solve the problem.

msarahan:
If you depend this much on build number, you're going to have a bad time.  You are giving it far more value than it actually has.  You need to reexamine what a version means, and trust that getting the latest build number (whatever it is) is almost certainly what you want.  Packaging should not change the functionality of software.  If it does, then the packagers are not doing a good job, and should honestly not be trusted.  Any software changes should be submitted upstream.  Build numbers exist simply because we don't always get things right on the first try.

Sorry, we really frown on meaninglessly bumping build number, because it means we have to rebuild software for no reason.  We are time-limited, like everyone else.  Please move on from this issue.

tinyheero:
That's unfortunate. I was just trying to fix a commit upstream of this that misused the build numbers in the first place. 

Anyways. Thanks for your time.

PR #631, opened 2016-03-10
Bump NumPy version in numpy-openblas recipe

bjodah:
1.10.0 on PyPI gives 404.


bjodah:
Regarding `build: string:` I am quite clueless to why differs from e.g. scipy-openblas?
cf. https://github.com/conda/conda-recipes/commit/acac86efe5af4a0d364e2d91c854f9b658e88024#comments

PR #641, opened 2016-03-16
Fix path errors

mutirri:
Fixing path issues according to this: https://github.com/conda/conda-recipes/pull/542


ccordoba12:
There are some packages that are failing to build on Linux and Mac (according to Anaconda build), but I think we should merge to get things back in order.

@kalefranz?

ccordoba12:
I'm merging this to fix the problems introduced by #542. We can look at the build failures later :-)

PR #642, opened 2016-03-16
python/raml package

kalefranz:


ilanschnell:
Is it possible to fun the tests in a cleaner way.  Something like:
```
# run_test.py
import raml
assert raml.test().wasSuccessful()
```

kalefranz:
Something like that @ilanschnell ?

ilanschnell:
Yeah

python/raml/bld.bat
```diff
@@ -0,0 +1,9 @@
+set INCLUDE=%LIBRARY_INC%
+set LIBPATH=%LIBRARY_LIB%
+set LIB=%LIBRARY_LIB%
+
+python setup.py install
+if errorlevel 1 exit 1
+
+copy %LIBRARY_BIN%\yaml.dll %SP_DIR%\raml\ext\
```
kalefranz: @ilanschnell note the `%SP_DIR%\raml\ext\` directory.
compare to: https://bitbucket.org/kalefranz/yaml/src/default/setup.py?fileviewer=file-view-default#setup.py-20

python/raml/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+  name: raml
+  version: 0.11.7
+
+build:
+  number: 0
+  # Windows has issues updating stuff if conda is using files itself. Copy rather than link.
+  no_link: .*\.(pyd|dll)        [win]
+
+source:
+  hg_url: ssh://hg@bitbucket.org/kalefranz/yaml
```
groutr: I think sources should be fetched via http/https instead of ssh.

PR #655, opened 2016-03-31
Fix incomplete manual merge in gsl recipe.

pavoljuhas:
Remove conflict marker.  Restore the `--with-pic` option from
the main branch.  Reformat comments to fit within 78 characters.


msarahan:
Thanks, good catch.

PR #658, opened 2016-04-02
Fix lapack for conda_builder_linux from ContinuumIO/docker-images

bjodah:
See https://github.com/ContinuumIO/docker-images/issues/23
ping @msarahan 

Actually I would need to remove both `gcc` and `libgcc` from meta.yaml for this to work out of the
box with `conda_builder_linux`.

What is the policy for adding compiler requirements in conda-recipes? My take is that
they can be kept in the recipe but commented out.

### DIFF
--- a/lapack/build.sh
+++ b/lapack/build.sh
@@ -1,8 +1,11 @@
 mkdir build
 cd build
 
+# CMAKE_INSTALL_LIBDIR="lib" suppresses CentOS default of lib64 (conda expects lib)
+
 cmake \
   -DCMAKE_INSTALL_PREFIX=$PREFIX \
+  -DCMAKE_INSTALL_LIBDIR="lib" \
   -DBUILD_TESTING=OFF \
   -DBUILD_SHARED_LIBS=ON \
   -DLAPACKE=ON \


msarahan:
Perfect, I think this is exactly what's needed.

gcc should be removed from meta.yaml, but libgcc might need to stay in.  The reason is that the docker build image contains runtimes that are potentially (probably) newer than system ones.  These are backwards compatible, but any code you compile with the docker image may not find the symbols it needs in older runtimes.

This libgcc is a bit of a work in progress.  If you need it now, grab it from my anaconda.org channel:

    conda install -c msarahan libgcc

The one on the defaults channel should soon be updated to be the same.  The tricky detail about doing things with GCC the way that we have is that our runtimes need to always be the same or newer version than the system libraries.

Ideally, compiler requirements could be made explicit, or at the very least, encoded into readily accessible metadata.  We sort of do that on Windows, since each version of Python uses a different version of Visual Studio.  This is another WIP, and the most current issue on the topic is at https://github.com/conda/conda-build/pull/848 - we'd welcome your thoughts on the topic.

bjodah:
@msarahan great! Thanks for the explanation and the hint about libgcc.

PR #663, opened 2016-04-13
muparser: require gcc libs

jschueller:


jschueller:
@msarahan, hello, any thoughts about this ?

msarahan:
Fine with me.  Any comment for history's sake on why it is necessary?  Did you run into trouble?

Also, consider submitting recipes to conda-forge instead of here.  Over time, we're going to be moving most/all of the recipes here elsewhere, and replacing the contents of this repo with submodules to "canonical sources" - either at package maintainer's repositories, or at conda-forge.

jschueller:
@msarahan. No trouble, it's just that I now build my packages on a chrooted environment where I do not install gcc on purpose to use the conda provided package only.

msarahan:
I'm going to merge this.  As time goes on, I think we're moving towards compiler packages as something that verify version, rather than something that provides actual compilers.  In tandem, we'll be pushing for adoption of standard build images, such as docker (https://hub.docker.com/r/continuumio/conda_builder_linux/)

The main drawback to shipping a compiler package is that it does not couple lower-level libraries (like glibc) to the build - so even when you use a "standard" compiler package, your package's dependencies vary depending on what system you build on.

All this just FYI.  No need to change anything now.

PR #664, opened 2016-04-13
nlopt: explicit python version dependency

jschueller:
apparently packages that depend on numpy should always list their python dependency explicitly:
https://github.com/conda/conda-build/issues/835

this fixes the build for another python version:
```conda build --python 3.4 python/nlopt```

### DIFF
--- a/python/nlopt/meta.yaml
+++ b/python/nlopt/meta.yaml
@@ -3,15 +3,16 @@ package:
   version: 2.4.2
 
 build:
-  number: 1
+  number: 2
 
 requirements: # [not win]
   build:      # [not win]
+    - python {{PY_VER}}* # [not win]
     - gcc     # [linux]
     - numpy   # [not win]
 
   run:
-    - python  # [not win]
+    - python {{PY_VER}}*  # [not win]
     - numpy   # [not win]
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: python/nlopt/meta.yaml [hunk 1]
```diff
@@ -5,19 +5,20 @@ package:
...
+requirements:
+  build:
+    - python {{PY_VER}}*
+    - numpy
     - gcc     # [linux]
-    - numpy   # [not win]
 
   run:
-    - python  # [not win]
-    - numpy   # [not win]
```
ccordoba12: Is this working for Windows?

jschueller:
@ccordoba12, I did not test.
Do you think it would be wrong to depend on python/numpy on windows ?

jschueller:
@ccordoba12, I see your point, Python bindings do not seem to be built on windows, I modified the comit so to change only the python dependency.

jschueller:
@dubourg this allows to upload openturns-1.7_py34!

dubourg:
Thanks J!

ccordoba12:
Thanks @jschueller! Merging :-)

PR #671, opened 2016-05-02
pin versions in r-packages.

groutr:
This should be automated, but conda-build doesn't have this capability right now.


groutr:
no longer relevant

PR #681, opened 2016-06-03
graphviz: Major changes

stuarteberg:
The current version of the `graphviz` package in the conda `defaults` channel seems to be broken.  For example, it expects the user to have `libtool`, `expat`, and `tk` installed to the system.  This PR fixes it.

Also, the graphviz website is (apparently) quite unreliable -- it was down earlier today, for example.  It seems be back up now, but the problem seems to [pop up with some frequency](https://github.com/ellson/graphviz/issues/77).  So now I've changed this recipe to clone from the official git repo instead of downloading a tarball.

- Build from scratch, from the official git repo
- Require libtool and expat (they seem to be necessary, the package is broken without them)
- Don't use cairo on OS X, since it's an optional dependency and it isn't available in the defaults channel (for OS X).
- Moved post-link code into build.sh

I tested these changes on Mac and Linux.  Packages available on my channel:

```
conda install -c stuarteberg graphviz
```

One caveat: As always, Graphviz cannot build without yacc (or bison) and lex (or flex), so you can't build this recipe without those tools installed to your system.  (They are not yet available as conda packages.)

cc: @jakirkham

### DIFF
--- a/graphviz/build.sh
+++ b/graphviz/build.sh
@@ -6,11 +6,26 @@ export CXXLAGS="${CFLAGS}"
 export CPPFLAGS="-I${PREFIX}/include"
 export LDFLAGS="-L${PREFIX}/lib"
 
+./autogen.sh
+
 if [ `uname` == Darwin ]; then
-./configure --prefix=$PREFIX --with-quartz | tee configure.log 2>&1
+./configure --prefix=$PREFIX --with-quartz
 else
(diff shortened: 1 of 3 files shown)


ccordoba12:
@stuarteberg, I took the first step to fix our many problems with Graphviz :-)

Please see:

https://github.com/conda-forge/staged-recipes/pull/568

(Now every fixes or new recipes go to conda-forge ;-)

PR #691, opened 2016-07-22
gcc-4.8: add library paths reported by system linker

svenpeter42:
Right now only part of the system's libc is symlinked into the conda env - namely the startup files crtXXX.o.
In addition the path to the actual shared object libc.so also needs to be added to the spec file in case a custom
linker is used inside the conda env (e.g. because binutils is requested).

This fixes the following error I get when I conda build a package that requires both gcc-4.8 and binutils:
`/home/speter/miniconda2/envs/_build/bin/ld: cannot find -lc
collect2: error: ld returned 1 exit status
Installation failed: gcc is not able to compile a simple 'Hello, World' program.`

 cc @stuarteberg

### DIFF
--- a/gcc-4.8/post-link.sh
+++ b/gcc-4.8/post-link.sh
@@ -46,7 +46,7 @@ else
     c_runtime_obj_files_found=0
 
     # Try locating crtXXX.o in default library search paths
-    for library_path in $(ld --verbose | grep SEARCH_DIR | sed -r 's/SEARCH_DIR\("=?([^"]*)"\);/ \1/g'); do
+    for library_path in $(/usr/bin/ld --verbose | grep SEARCH_DIR | sed -r 's/SEARCH_DIR\("=?([^"]*)"\);/ \1/g'); do
         for obj_file in $C_RUNTIME_OBJ_FILES; do
             obj_file_full_path="$library_path/$obj_file"
             if [[ -e "$obj_file_full_path" ]]; then
--- a/gcc-4.8/post-link.sh
+++ b/gcc-4.8/post-link.sh
@@ -111,6 +111,13 @@ else
         # ... yada yada ... -isystem ${INCDIR}
         sed -i ':a;N;$!ba;s|\(*cpp:\n[^\n]*\)|\1 -isystem '${INCDIR}'|g' "${SPECS_FILE}"
     done
+
+    #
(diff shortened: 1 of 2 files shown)


gcc-4.8/post-link.sh
```diff
@@ -111,6 +111,13 @@ else
         # ... yada yada ... -isystem ${INCDIR}
         sed -i ':a;N;$!ba;s|\(*cpp:\n[^\n]*\)|\1 -isystem '${INCDIR}'|g' "${SPECS_FILE}"
     done
+
+    #
+    # Linux Portability Issue #2.5: linker also needs to find the rest of libc (i.e. libc.so in addition to crtXXX.o)
+    #
+    for library_path in $(/usr/bin/ld --verbose | grep SEARCH_DIR | sed -r 's/SEARCH_DIR\("=?([^"]*)"\);/ \1/g'); do
```
stuarteberg: Using the full path to the system `/usr/bin/ld` instead of `ld` is a good idea. (It avoids problems if the user has a custom install of binutils.)  While you're at it, can you also fix the call to `ld` on line 49, above?

stuarteberg:
You should bump the build number.

stuarteberg:
LGTM.

stuarteberg:
@frol: Just in case it matters for your gcc-5 package -- @svenpeter42 discovered that the gcc-4.8 `post-link.sh` script doesn't work properly if the user has also installed `binutils` via conda.  This PR fixes it, so I guess it should also be added to gcc-5, if you're still maintaining that recipe.  (At the moment, there is no "official" binutils package from conda, so this will only affect a very small minority of users.)

If you're not still maintaining the gcc-5 recipe, then sorry for the spam. :-)

ccordoba12:
Pinging @mingwandroid about this one.

frol:
Thank you for pinging me! I do still maintain `gcc-5` package as well as I maintain `conda-forge/binutils` feedstock.

ostrokach:
Is anything holding back this PR? I just spent half a day tring to figure out why conda's `gcc` / `mpich` was giving me an error `ld: cannot find -lc`:

```
$ mpicc -o mpi_hello_world mpi_hello_world.c
/home/{username}/anaconda/envs/{projectname}/bin/ld: cannot find -lc
collect2: error: ld returned 1 exit status
```

unless I specified `-L/usr/lib/x86_64-linux-gnu`.

PR #705, opened 2016-12-14
fix bash specific comparitor in patch of ldpaths.in

stevecassidy:
comparison with == makes this script bash specific and causes an error on Ubuntu where /bin/sh is dash. This change should fix this issue.


mingwandroid:
Thanks.

PR #707, opened 2017-01-19
adding various r-packages

quasiben:
cc @mingwandroid

### DIFF
--- a/r-packages/r-anomalydetection/bld.bat
+++ b/r-packages/r-anomalydetection/bld.bat
@@ -1,0 +1,8 @@
+"%R%" CMD INSTALL --build .
+if errorlevel 1 exit 1
+
+@rem Add more build steps here, if they are necessary.
+
+@rem See
+@rem http://docs.continuum.io/conda/build.html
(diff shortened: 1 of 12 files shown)


mingwandroid:
We shouldn't see any .template files here, conda-build autogenerated those, I guess you got them from downloaded packages' info dirs?

Can you use `conda-skeleton cran` to make the recipes instead (though you need to change the case of various license families after unfortunately).

Also, I see various packages depending on `r`, they should depend on `r-base` instead and any other packages they happen to depend upon. It seems those packages were hand-crafted, perhaps?

The reason I'm need to be picky about this is that when I update to the latest releases, I use `conda skeleton cran` to generate new ones and then eyeball every single package for differences and I need that to be as easy as possible since there are now ~300 packages.

quasiben:
@mingwandroid pulled from cran and rebased

mingwandroid:
I built and uploaded all of these now.

quasiben:
thanks @mingwandroid !

PR #710, opened 2017-02-15
New gcc recipes

msarahan:
For native gcc, change CPU -march from westmere (i3/i5/i7) to nocona (core 2).


msarahan:
Per discussion with @mingwandroid, merging.

PR #721, opened 2017-06-02
Allowing dynamic java for r-base (breaks rJava)

Alexandrescu:
R is not linking to the correct java JVM.


mingwandroid:
In what way is following a symlink to the specific location on your computer at build-time dynamic in a generally useful sense to anyone but you?

PR #730, opened 2017-06-15
gsl: build with system compiler.

pavoljuhas:
Avoid linking with possibly non-compatible shared libraries from
conda "gcc" or "libgcc" packages.  Using system compiler keeps
GSL compatible with other packages built with standard system
tools.

This partially reverts 40e929cee7f8453d173818d36420879706046c32.

### DIFF
--- a/gsl/meta.yaml
+++ b/gsl/meta.yaml
@@ -8,14 +8,12 @@ source:
   md5: d8f70abafd3e9f0bae03c52d1f4e8de5
 
 build:
-  number: 1
+  number: 2
 
 requirements:
-  build:
-    - gcc
-
-  run:
-    - libgcc
+  # Build gsl with system compiler.  Do not use conda "gcc" nor
+  # "libgcc" packages as they may install shared libraries that
+  # are incompatible with their system version.
 
 about:
   home: http://www.gnu.org/software/gsl/


mingwandroid:
We aim to be compatible with the `Anaconda Distribution`, not compatible with your system's distribution.

pavoljuhas:
I feel using the system Xcode compiler and system libraries is more likely to produce packages compatible across different OS X versions.  It seems easier to make GSL and boost compatible by dropping the libgcc runtime dependency for the "gsl" package rather than rebuilding the whole Anaconda with its own gcc.  At any rate, GSL should not install dependencies that it does not need.

As it is the Anaconda c++ compiler behaves in a bit confusing way as to which libstdc++ gets used:
```sh
$ conda create -n bld gcc
$ source activate bld
$ echo "int main() { return 0; }" > t.cpp
$ c++ t.cpp
$ otool -L a.out
a.out:
	@rpath/libstdc++.6.dylib (compatibility version 7.0.0, current version 7.19.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1213.0.0)
	/Users/pjuhas/arch/anaconda/envs/bld/lib/libgcc_s.1.dylib (compatibility version 1.0.0, current version 1.0.0)

# however a.out does not have any RPATH defined thus execution of a.out
# actually loads the system libstdc++
$ DYLD_PRINT_LIBRARIES=1 ./a.out
dyld: loaded: /Users/pjuhas/tmp/./a.out
dyld: loaded: /usr/lib/libstdc++.6.dylib
...
```

The `a.out` was linked with conda libstdc++.6.dylib but runs with the system version.  I guess in some cases those libraries can be switched one for another, but they are not fully compatible as shown by test failure for my package that links with gsl and boost (both from the conda defaults channel) -
 https://travis-ci.org/pavoljuhas/libdiffpy/jobs/243311738#L523.

mingwandroid:
For now, until I get time to look into this properly, can you build your own GSL and put it on a channel on anaconda.org?

pavoljuhas:
obsolete

